PR #17, opened 2023-01-11
Respond to HEAD requests to /blobs/:addr

shinypb:
Sending a HEAD request to the blob retrieval endpoint now gives you a succinct way of discovering whether a given blob has been stored before or not: 200 if it has, 404 if it hasn't. This lets us avoid sending a blob up to the storage system when it's already been stored.

I also added a few minor niceties to pounce while working on this:
- default to using the filename of the wasm binary rather than "unnamed"
- show the size of the binary + input payload in the output so people aren't surprised when big requests are slow to start (my 37+ MB object detection demo takes a solid 5–10 seconds, at least)

### DIFF
--- a/pounce/src/main.rs
+++ b/pounce/src/main.rs
@@ -12,6 +12,7 @@ use anyhow::{anyhow, Result};
 use clap::{Parser, Subcommand};
 use mdns_sd::{ServiceDaemon, ServiceEvent};
 use owo_colors::OwoColorize;
+use thousands::Separable;
 use uuid::Uuid;
 
 use std::fs::File;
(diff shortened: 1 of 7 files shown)


pounce/src/main.rs
```diff
@@ -113,19 +114,31 @@ fn run(
...
 
-    let name = name.unwrap_or_else(|| "unnamed".to_string());
+    let name = name.unwrap_or_else(|| {
+        // use the filename component of binarypath, e.g. /foo/bar.wasm -> bar.wasm
+        binarypath
+            .file_name()
+            .and_then(|name| name.to_str())
+            .map(|z| z.to_string())
+            .unwrap_or_else(|| "unnamed".to_string())
+    });
```
ceejbot, line 134: This is a nicely thoughtful addition!

pounce/src/main.rs
```diff
@@ -113,19 +114,31 @@ fn run(
...
     let description = description.unwrap_or_else(|| "posted via command-line".to_string());
 
-    println!("Sending {} to serval agent...", name.blue().bold());
+    println!(
+        "Sending {} ({} bytes for binary + payload) to serval agent...",
+        name.blue().bold(),
+        binary_payload_size.separate_with_commas(),
+    );
```
ceejbot, line 141: Nice again!

utils/src/blobs.rs
```diff
@@ -47,6 +47,16 @@ impl BlobStore {
         Ok(stream)
     }
 
+    // Given a content address, determine whether we have a blob stored there or not.
+    pub async fn has_blob(&self, address: &str) -> Result<bool, ServalError> {
```
ceejbot, line 51: This is a solid addition to the API!

pounce/src/main.rs
```diff
@@ -113,19 +114,31 @@ fn run(
...
+            .map(|z| z.to_string())
+            .unwrap_or_else(|| "unnamed".to_string())
+    });
     let description = description.unwrap_or_else(|| "posted via command-line".to_string());
 
-    println!("Sending {} to serval agent...", name.blue().bold());
+    println!(
+        "Sending {} ({} bytes for binary + payload) to serval agent...",
+        name.blue().bold(),
+        binary_payload_size.separate_with_commas(),
```
ceejbot, line 140: We'll eventually want to use crates like `humansize` to turn byte counts into more human-readable sizes, but this is a solid improvement!

ceejbot (approved):
🐑

PR #18, opened 2023-01-11
serval-agent - allow an instance to run without offering blob storage

shinypb:
Right now, our blob storage is a temporary implementation that just stores data in $TMPDIR. If
multiple copies of the agent are running on different machines, that means that blob storage
won't be reliable -- if you don't happen to ask an instance on the correct box, you might not
get your data back.

To avoid this, let's make it so you can run an instance of the agent without it implementing
any blob storage stuff. Changes required to do this:

1. Add command line parsing via clap and allow people to specify whether this instance should be
a storage node or not. Rather than a true/false switch, I've made this auto/never/always, with
the assumption that auto will someday make some judgements about whether the current machine is
a viable platform upon which to run the storage system. e.g. an embedded device without much
disk space might decide not to run the storage server.

2. Refactor `RunnerState` so that it holds an `Option<BlobStore>` instead of a `BlobStore`. This
involved some yak-shaving to make the ::new function return a `Result`.

3. Change blob-related API endpoints to return 503 errors if you try to hit them when storage is
disabled.

Future changes required but not done here:
- API calls to /blobs/... should proxy to a node that is advertising the serval_storage role,
rather than just returning a 503 error.

While I was was in here, I modified `BlobStore::new` function to ensure that the path it has been
given exists and points to a writable directory. If the path does not exist, we try to create it.
I also changed the default path from `/tmp` to `/tmp/serval_storage`.

Phew!

### DIFF
--- a/serval-agent/src/api/jobs.rs
+++ b/serval-agent/src/api/jobs.rs
@@ -133,7 +133,13 @@ pub async fn run_stored_job(
     Path(blob_addr): Path<String>,
 ) -> impl IntoResponse {
     let locked = state.lock().await;
-    let Ok(binary) = locked.storage.get_bytes(&blob_addr).await else {
+
+    let Some(storage) = locked.storage.as_ref() else {
+        // todo: in this case, we should proxy this request to another node that is advertising the serval_storage role
+        return (StatusCode::SERVICE_UNAVAILABLE, "Storage is not available").into_response();
+    };
+
+    let Ok(binary) = storage.get_bytes(&blob_addr).await else {
         return (
             StatusCode::NOT_FOUND,
             format!("Blob {} not found", &blob_addr),
--- a/serval-agent/src/api/mod.rs
+++ b/serval-agent/src/api/mod.rs
@@ -7,7 +7,7 @@ use axum::{
 use http::header::HeaderValue;
 use serde::{Deserialize, Serialize};
 use tokio::sync::Mutex;
-use utils::blobs::BlobStore;
+use utils::{blobs::BlobStore, errors::ServalError};
 use uuid::Uuid;
 
 use std::sync::Arc;
@@ -18,20 +18,24 @@ pub mod storage;
 
 #[derive(Debug, Clone, Serialize)]
 pub struct RunnerState {
-    storage: BlobStore,
+    storage: Option<BlobStore>,
     jobs: HashMap<String, JobMetadata>,
     total: usize,
     errors: usize,
 }
 
 impl RunnerState {
-    pub fn new(blob_path: PathBuf) -> Self {
-        RunnerState {
-            storage: BlobStore::new(blob_path),
+    pub fn new(blob_path: Option<PathBuf>) -> Result<Self, ServalError> {
+        let storage = match blob_path {
+            Some(path) => Some(BlobStore::new(path)?),
+            None => None,
+        };
+        Ok(RunnerState {
+            storage,
             total: 0,
             errors: 0,
             jobs: HashMap::new(),
-        }
+        })
     }
 }
 
--- a/serval-agent/src/api/mod.rs
+++ b/serval-agent/src/api/mod.rs
@@ -10,8 +10,8 @@ use tokio::sync::Mutex;
 use utils::blobs::BlobStore;
 use uuid::Uuid;
 
-use std::collections::HashMap;
 use std::sync::Arc;
+use std::{collections::HashMap, path::PathBuf};
 
 pub mod jobs;
 pub mod storage;
@@ -25,9 +25,9 @@ pub struct RunnerState {
 }
 
 impl RunnerState {
-    pub fn new(blob_path: String) -> Self {
+    pub fn new(blob_path: PathBuf) -> Self {
         RunnerState {
-            storage: BlobStore::new(blob_path.into()),
+            storage: BlobStore::new(blob_path),
             total: 0,
             errors: 0,
             jobs: HashMap::new(),
--- a/serval-agent/src/api/storage.rs
+++ b/serval-agent/src/api/storage.rs
@@ -16,7 +16,12 @@ pub async fn get_blob(
     // Yeah, I don't like this.
     let state = state.lock().await;
 
-    match state.storage.get_stream(&blob_addr).await {
+    let Some(storage) = state.storage.as_ref() else {
+        // todo: in this case, we should proxy this request to another node that is advertising the serval_storage role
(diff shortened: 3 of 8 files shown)


serval-agent/src/api/storage.rs
```diff
@@ -16,7 +16,12 @@ pub async fn get_blob(
     // Yeah, I don't like this.
     let state = state.lock().await;
 
-    match state.storage.get_stream(&blob_addr).await {
+    let Some(storage) = state.storage.as_ref() else {
+        // todo: in this case, we should proxy this request to another node that is advertising the serval_storage role
+        return (StatusCode::SERVICE_UNAVAILABLE, "Storage not available").into_response();
```
shinypb, line 21: I'd love to just be able to do something like:

````rust
return proxy_request_to_role("serval_storage", format!("/blob/{blob_addr}"));
````

It'd be even cooler if there was an easy way to get the current request we're in the middle of serving and pass it along to that hypothetical `proxy_request_to_role` function.

File: serval-agent/src/api/mod.rs [hunk 2]
```diff
@@ -7,31 +7,35 @@ use axum::{
...
     total: usize,
     errors: usize,
 }
 
 impl RunnerState {
-    pub fn new(blob_path: String) -> Self {
-        RunnerState {
-            storage: BlobStore::new(blob_path.into()),
+    pub fn new(blob_path: Option<PathBuf>) -> Result<Self, ServalError> {
+        let storage = match blob_path {
```
shinypb, line 29: If there's a more idiomatic way to do this, I'd love to hear it. I tried variants along these lines...

````rust
let storage = blob_path.map(|path| BlobStore::new(path)?);
````

...but `Option` and `Result` don't mix super well.

File: serval-agent/src/api/jobs.rs [hunk 3]
```diff
@@ -133,7 +133,13 @@ pub async fn run_stored_job(
     Path(blob_addr): Path<String>,
 ) -> impl IntoResponse {
     let locked = state.lock().await;
-    let Ok(binary) = locked.storage.get_bytes(&blob_addr).await else {
+
+    let Some(storage) = locked.storage.as_ref() else {
+        // todo: in this case, we should proxy this request to another node that is advertising the serval_storage role
+        return (StatusCode::SERVICE_UNAVAILABLE, "Storage is not available").into_response();
+    };
```
ceejbot, line 140: Yeah, I can see we'd implement redirection right here.

serval-agent/src/api/mod.rs
```diff
@@ -7,31 +7,35 @@ use axum::{
...
+            Some(path) => Some(BlobStore::new(path)?),
```
ceejbot, line 30: Fallibility is infectious. (A comment intended to indicate that I have seen this change and concur.)

serval-agent/src/api/storage.rs
ceejbot: Noted! This should be quite possible.

serval-agent/src/main.rs
```diff
@@ -14,29 +14,54 @@ use axum::{
...
+    Auto,
+    Never,
+    Always,
+}
+
+#[derive(Parser, Debug)]
+struct Args {
+    #[clap(long, value_enum, default_value = "auto")]
+    // Whether to assume the "storage" role in the mesh; defaults to StorageRoleConfig::Auto
+    storage_role: StorageRoleConfig,
```
ceejbot, line 39: And you will  have noted how nicely clap handles enums as option values! (Clap is v good.)

serval-agent/src/main.rs
```diff
@@ -14,29 +14,54 @@ use axum::{
...
-    let blob_path = std::env::var("BLOB_STORE").unwrap_or_else(|_| "/tmp".to_string());
+    let blob_path = match args.storage_role {
+        StorageRoleConfig::Never => None,
+        _ => Some(
+            std::env::var("BLOB_STORE")
+                .map(PathBuf::from)
+                .unwrap_or_else(|_| std::env::temp_dir().join("serval_storage")),
+        ),
+    };
+    let should_advertise_storage = blob_path.is_some();
```
ceejbot, line 61: My one comment here is that I'm still in the 12-factor-app camp and prefer env vars to push config through via OS-level daemon runners, but this is not a preference we should fret about right now. Landing the functionality and continuing to move forward is more important. But I wanted to mention it as something we'll eventually need to resolve one way or the other.

serval-agent/src/main.rs
```diff
@@ -14,29 +14,54 @@ use axum::{
...
 
-    log::info!("serval agent blob store mounted; path={blob_path}");
-    let state = Arc::new(Mutex::new(RunnerState::new(blob_path)));
+    log::info!("serval agent blob store mounted; path={blob_path:?}");
```
ceejbot, line 63: Actual PR suggestion: we should conditionalize the entire log line instead of debug-printing the option. If we are not advertising storage, say so. If we are, say so and give the location.

serval-agent/src/main.rs
```diff
@@ -55,6 +80,11 @@ async fn main() -> Result<()> {
     log::info!("serval agent daemon listening on {}", &addr);
 
     advertise_service("serval_daemon", port, None)?;
+    if should_advertise_storage {
+        advertise_service("serval_storage", port, None)?;
```
ceejbot, line 84: oh hey speaking of log lines, that line can pop down here and I get what I dream of!

ceejbot (approved):
One tiny suggestion but really, :sheep: it.

serval-agent/src/main.rs
shinypb: I've gone back and forth on this a few times and I welcome future iterations from you. I'm about to push a commit that shuffles a bit of code around and I mostly-like where it's ended up:

````rust
if blob_path.is_some() {
    log::info!("serval agent blob store mounted; path={blob_path:?}");
    advertise_service("serval_storage", port, None)?;
}
````

This keeps the logging right next to the advertisement. I don't have a log statement for the negative case, but I think that's ok; there's already a log statement that happens when the advertisement occurs, so it's pretty easy to tell from a set of logs whether we were doing storage or not:

````
[2023-01-12T17:26:39Z INFO  serval_agent] serval agent blob store mounted; path=Some("/var/folders/1v/50bp7sys29s1cxjxyn03kbqw0000gn/T/serval_storage")
[2023-01-12T17:26:39Z INFO  utils::mdns] Advertising serval_storage; domain=_serval_storage._tcp.local. port=8100 props=None
````

PR #19, opened 2023-01-12
Introduce proxy_unavailable_services middleware

shinypb:
This adds a new middleware function that looks at the path of the incoming request and rejects any under the `/storage/` namespace if we are running with storage disabled.

In the not-so-distant future, this middleware should instead proxy the request to a node that is advertising the `serval_storage` role. I will begin working on that in a follow-up branch.

### DIFF
--- a/serval-agent/src/api/storage.rs
+++ b/serval-agent/src/api/storage.rs
@@ -15,11 +15,7 @@ pub async fn get_blob(
 ) -> impl IntoResponse {
     // Yeah, I don't like this.
     let state = state.lock().await;
-
-    let Some(storage) = state.storage.as_ref() else {
-        // todo: in this case, we should proxy this request to another node that is advertising the serval_storage role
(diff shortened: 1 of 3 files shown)


serval-agent/src/api/storage.rs
```diff
@@ -15,11 +15,7 @@ pub async fn get_blob(
 ) -> impl IntoResponse {
     // Yeah, I don't like this.
     let state = state.lock().await;
-
-    let Some(storage) = state.storage.as_ref() else {
-        // todo: in this case, we should proxy this request to another node that is advertising the serval_storage role
-        return (StatusCode::SERVICE_UNAVAILABLE, "Storage not available").into_response();
-    };
+    let storage = state.storage.as_ref().unwrap();
```
shinypb, line 18: “I'll just put this over here with the rest of the don't-like-this.”

the-it-crowd-moss

ceejbot (approved):
Excellent use of the middleware!

PR #21, opened 2023-01-13
Implement IntoResponse for ServalError

ceejbot:
This cleans up a lot of the http machinery, at the expense of some specific logging. The logging does include the error text, so we can improve that if we want the details back.


shinypb (approved):
I <3 this

PR #22, opened 2023-01-13
Proxy storage requests to another node if we aren't in the storage role

shinypb:
This expands the `proxy_unavailable_services` stub that we added earlier in the week to actually implement proxying, rather than returning a 503 error. When we get a request under the `/storage/` namespace, we now use mdns to look for a node advertising the `_serval_storage` role, forward a copy of the request to them, and pass the result back. With a handful of exceptions, all of the headers from the original are passed through unchanged, so non-trivial requests that make use of things like cookies should theoretically Just Work™.

There's no retry logic if things don't work for whatever reason, but I think that's ok (especially at this stage). If anything goes wrong trying to proxy the request to another node, it's surfaced to the original requestor as a [503 (service unavailable)](https://http.cat/503) error.

## Testing

:one: Spin up two instances, making one the storage node and one not:

- `PORT=8101 STORAGE_ROLE=always cargo run --bin serval-agent`
- `PORT=8100 STORAGE_ROLE=never cargo run --bin serval-agent`

(I'm hard-coding the port numbers here so the order I start these instances up doesn't change which port number is storage vs. not.)

:two: Make a request to the storage node

````sh
$ curl -D - localhost:8101/storage/blobs -X PUT -sT README.md
HTTP/1.1 100 Continue

HTTP/1.1 200 OK
content-type: text/plain; charset=utf-8
content-length: 64
x-clacks-overhead: GNU/Terry Pratchett
date: Fri, 13 Jan 2023 21:41:01 GMT

7781ffbc12ffcec30129d37c63e0576a34c14a39d2ff71a42f2848948b7e9c61%                                                                                              
````

In particular, note the lack of a `serval-proxied-from` header in the response.

:three: Make a request to the non-storage node

````sh
$ curl -D - localhost:8100/storage/blobs -X PUT -sT README.md
HTTP/1.1 200 OK
content-type: text/plain; charset=utf-8
content-length: 64
x-clacks-overhead: GNU/Terry Pratchett
date: Fri, 13 Jan 2023 21:40:53 GMT
serval-proxied-from: <todo: put instance_id from `info`'s properties here>

e3b0c44298fc1c149afbf4c8996fb92427ae41e4649b934ca495991b7852b855%
````

Behold the gentle glory of the `serval-proxied-from` header, and if you look at the log output of the non-storage node, you'll see this:
````
[2023-01-13T22:22:37Z INFO  serval_agent::api::proxy] proxy_unavailable_services intercepting request; path=/storage/blobs
````

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -17,6 +17,7 @@ dotenvy = "0.15.6"
 env_logger = "0.10.0"
 log = "0.4.17"
 mdns-sd = { version = "0.5.9", default-features = false, features = ["async"] }
+reqwest = { version = "0.11.13",  default-features = false, features = ["brotli", "json", "rustls"] }
 serde = { version = "1.0.149", features = ["derive"] }
 serde_json = "1.0.89"
 tokio = { version = "1.22.0", features = ["full"] }
--- a/agent/src/api/mod.rs
+++ b/agent/src/api/mod.rs
@@ -1,9 +1,8 @@
 use anyhow::Result;
 use axum::{
-    extract::State,
     http::{Request, StatusCode},
     middleware::Next,
-    response::{IntoResponse, Response},
+    response::Response,
 };
 use http::header::HeaderValue;
 use serde::{Deserialize, Serialize};
@@ -15,6 +14,7 @@ use std::sync::Arc;
 use std::{collections::HashMap, path::PathBuf};
 
 pub mod jobs;
+pub mod proxy;
 pub mod storage;
 
 #[derive(Debug, Clone, Serialize)]
@@ -98,26 +98,4 @@ pub async fn clacks<B>(req: Request<B>, next: Next<B>) -> Result<Response, Statu
 /// Respond to ping. Useful for monitoring.
 pub async fn ping() -> String {
     "pong".to_string()
-}
-
-pub async fn proxy_unavailable_services<B>(
-    State(state): State<AppState>,
-    req: Request<B>,
-    next: Next<B>,
-) -> Result<Response, StatusCode> {
-    if req.uri().path().starts_with("/storage/") {
-        let state = state.lock().await;
-        if state.storage.is_none() {
-            log::info!(
-                "proxy_unavailable_services intecepting request; path={}",
-                req.uri().path()
-            );
-            // todo: In this case, we should proxy this request to another node that is advertising
-            // the serval_storage role. For now, let's just barf.
-            return Ok((StatusCode::SERVICE_UNAVAILABLE, "Storage not available").into_response());
-        }
-    }
-
-    let response = next.run(req).await;
-    Ok(response)
 }
--- a/agent/src/api/mod.rs
+++ b/agent/src/api/mod.rs
@@ -86,10 +86,12 @@ pub struct Envelope {
 /// Remember what is important.
 pub async fn clacks<B>(req: Request<B>, next: Next<B>) -> Result<Response, StatusCode> {
     let mut response = next.run(req).await;
-    response.headers_mut().append(
-        "X-Clacks-Overhead",
-        HeaderValue::from_static("GNU/Terry Pratchett"),
-    );
+    if !response.headers().contains_key("X-Clacks-Overhead") {
+        response.headers_mut().append(
+            "X-Clacks-Overhead",
+            HeaderValue::from_static("GNU/Terry Pratchett"),
+        );
+    }
     Ok(response)
 }
 
--- a/agent/src/api/proxy.rs
+++ b/agent/src/api/proxy.rs
@@ -21,7 +21,7 @@ pub async fn proxy_unavailable_services<B>(
         let state = state.lock().await;
         if state.storage.is_none() {
             log::info!(
-                "proxy_unavailable_services intecepting request; path={}",
+                "proxy_unavailable_services intercepting request; path={}",
                 path,
             );
 
--- a/agent/src/api/proxy.rs
+++ b/agent/src/api/proxy.rs
@@ -125,3 +125,55 @@ async fn reqwest_response_to_axum_response(
 
     Ok(axum_resp)
 }
+
+#[cfg(test)]
+mod tests {
+
+    use anyhow::anyhow;
+    use axum::body::Bytes;
+    use http::response::Builder;
+    use reqwest::Response;
+    use utils::futures::get_future_sync;
+
+    use super::*;
+
+    fn get_axum_body_as_bytes(resp: axum::response::Response) -> Result<Bytes> {
+        let body = resp.into_body();
+        let Some(body_bytes) = get_future_sync(hyper::body::to_bytes(body)).ok() else {
+            return Err(anyhow!("Could not get body bytes"));
+        };
+
+        Ok(body_bytes)
+    }
+
+    #[test]
+    fn test_reqwest_response_to_axum_response() {
+        let mut reqwest_resp = Response::from(
+            Builder::new()
+                .status(418)
+                .body("<whistling noises intensify>")
+                .unwrap(),
+        );
+        reqwest_resp
+            .headers_mut()
+            .append("foo", HeaderValue::from_str("bar").unwrap());
+
+        // Make sure the Reqwest response matches expectations
+        assert_eq!(reqwest_resp.status(), 418);
+
+        // Make sure the conversion works
+        let result = get_future_sync(reqwest_response_to_axum_response(reqwest_resp));
+        assert!(result.is_ok());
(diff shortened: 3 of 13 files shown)

### CODE REVIEW

File: agent/src/api/mod.rs [hunk 1]
```diff
@@ -86,36 +86,16 @@ pub struct Envelope {
 /// Remember what is important.
 pub async fn clacks<B>(req: Request<B>, next: Next<B>) -> Result<Response, StatusCode> {
     let mut response = next.run(req).await;
-    response.headers_mut().append(
-        "X-Clacks-Overhead",
-        HeaderValue::from_static("GNU/Terry Pratchett"),
-    );
+    if !response.headers().contains_key("X-Clacks-Overhead") {
```
shinypb, line 89: Prior to this, proxied requests were ending up with two copies of this in their responses (one from each node that was involved). It's possible that I could've solved this else-how by swapping the order of the two `.route_layer` calls that install our middleware, but I am not sure that's any more grokkable. Happy to do this elsehow if you prefer.

agent/src/api/proxy.rs
```diff
@@ -0,0 +1,132 @@
...
+        }
+        inner_req = inner_req.header(k, v);
+    }
+    inner_req = inner_req.header(
+        "Serval-Proxied-For",
+        "<todo: put instance_id here once it exists in AppState>",
+    );
+
+    // Actually send the request
+    inner_req
```
shinypb, line 85: This whole block of code is a bit much; I am open to readability feedback.

agent/src/api/proxy.rs
```diff
@@ -0,0 +1,132 @@
...
+            );
+            resp
+        })
+        .map_err(|err| {
+            log::warn!("Failed to proxy request to other node; node={host}:{port}; err={err:?}");
+            err
+        })
+}
+
+async fn reqwest_response_to_axum_response(
```
shinypb, line 105: Not sure what sort of naming convention might be relevant here, but feedback is welcome. I really wish one could implement `IntoResponse` for `reqwest::Response`, but AFAIK that is not a thing in Rust.

agent/src/main.rs
```diff
@@ -61,6 +62,7 @@ async fn main() -> Result<()> {
         ),
     };
 
+    // todo: add instance_id (a UUID) to RunnerState
```
shinypb, line 65: Instance ID stuff intentionally left for another day

utils/src/mdns.rs
```diff
@@ -35,3 +37,51 @@ pub fn advertise_service(
 
     Ok(())
 }
+
+pub async fn discover_service(service_name: &str) -> Result<ServiceInfo, ServalError> {
+    discover_service_with_timeout(service_name, Duration::from_secs(30)).await
+}
+
+pub async fn discover_service_with_timeout(
```
shinypb, line 45: I quite like how this function turned out and am a little bit proud of it.

agent/src/api/proxy.rs
ceejbot: @chrisdickinson suggests the [newtype idiom](https://doc.rust-lang.org/rust-by-example/generics/new_types.html) as a pattern for this, though it might be overkill here. There are rules about implementing traits on types you don't own, as in, you can't. So you work around it by making a type that you _do_ own and implementing the trait on that:

```rust
pub struct ResponseInterchange(reqwest::Response);

impl IntoResponse for ResponseInterchange {
    ...
}
```

and then we can use `.into_response()` on our renamed reqwest response whenever we pull this trait implementation into scope.

cli/Cargo.toml
```diff
@@ -15,4 +15,6 @@ owo-colors = "3.5.0"
 reqwest = { version = "0.11.13", features = ["blocking", "brotli", "json", "multipart", "rustls"] }
 serde_json.workspace = true
 thousands = "0.2.0"
-uuid.workspace = true
+tokio.workspace = true
+utils = { path = "../utils" }
+uuid = { workspace = true }
```
ceejbot, line 20: le gasp on the curly braces (truly I do not care but it's on brand for me to make noise about this (my brand is bad))

cli/src/main.rs
```diff
@@ -218,26 +221,23 @@ fn blocking_maybe_discover_service_url(
...
+        return Err(anyhow!(format!(
+            "Failed to discover {service_type} node on the local network"
+        )));
+    };
+
+    let Some(addr) = info.get_addresses().iter().next() else {
+        // this should not ever happen, but computers
+        return Err(anyhow!(format!(
+            "Discovered a node that has no addresses",
+        )));
```
ceejbot, line 236: yeah wow, pathological case here

utils/src/errors.rs
```diff
@@ -66,6 +72,8 @@ impl IntoResponse for ServalError {
             ServalError::BlobAddressInvalid(_) => StatusCode::BAD_REQUEST,
             ServalError::BlobAddressNotFound(_) => StatusCode::NOT_FOUND,
             ServalError::IoError(_) => StatusCode::NOT_FOUND,
+            ServalError::ServiceNotFound => StatusCode::NOT_FOUND,
+            ServalError::ReqwestError(_) => StatusCode::INTERNAL_SERVER_ERROR,
```
ceejbot, line 76: Yeah, this is working out nicely-- no need to revisit the handlers every time we add an error type unless we *want* to handle one differently.

utils/src/mdns.rs
```diff
@@ -35,3 +37,51 @@ pub fn advertise_service(
...
+    let discover_service = async {
+        while let Ok(event) = receiver.recv_async().await {
+            let ServiceEvent::ServiceResolved(info) = event else {
+                // We don't care about other events here
+                continue;
+            };
+            if info.get_addresses().is_empty() {
+                // This should never happen, but let's check here so all consumer code can just
+                // info.get_addresses().get(0).upwrap() without needing to worry about it exploding.
+                continue;
```
ceejbot, line 66: early return/continue considered readable!

utils/src/mdns.rs
```diff
@@ -35,3 +37,51 @@ pub fn advertise_service(
...
+            return Ok(info);
+        }
+        Err(ServalError::ServiceNotFound)
+    };
+
+    let Ok(resp) = tokio_timeout(timeout_duration, discover_service).await else {
+        return Err(ServalError::ServiceNotFound);
+    };
+
+    resp
```
ceejbot, line 86: ✨

ceejbot (approved):
I offer a pattern for the reqwest -> axum response conversion, which you can do in this PR or follow up with, as you prefer!

agent/src/api/mod.rs
```diff
@@ -86,36 +86,16 @@ pub struct Envelope {
...
+        response.headers_mut().append(
+            "X-Clacks-Overhead",
+            HeaderValue::from_static("GNU/Terry Pratchett"),
```
jkbecker, line 92: 😹

cli/Cargo.toml
shinypb: Not even sure how this happened, tbh

Cargo.toml
```diff
@@ -17,6 +17,7 @@ dotenvy = "0.15.6"
 env_logger = "0.10.0"
 log = "0.4.17"
 mdns-sd = { version = "0.5.9", default-features = false, features = ["async"] }
+reqwest = { version = "0.11.13",  default-features = false, features = ["brotli", "json", "rustls"] }
```
ceejbot, line 20: btw I enabled brotli here because I was planning on compressing some requests to the agents; I would need to either find or write a brotli compression middleware for axum to make this plan reality. Or use something less new-fangled.

agent/src/api/proxy.rs
```diff
@@ -0,0 +1,179 @@
...
+    if path.starts_with("/storage/") {
+        let state = state.lock().await;
+        if state.storage.is_none() {
+            log::info!(
+                "proxy_unavailable_services intercepting request; path={}",
+                path,
+            );
+
+            let Ok(resp) = proxy_request_to_service(&req, "_serval_storage").await else {
+                // Welp, not much we can do
```
ceejbot, line 29: `log::warn!` here imo

ceejbot (commented):
See slack convo for a realization I just had.

agent/src/api/proxy.rs
shinypb: The warning logging happens inside of `proxy_request_to_service`: https://github.com/servals/serval-mesh/pull/22/files#diff-db7f2a5ed8d81b6f7a436d9e0117a6efde8fdb10389488c9f443c7b77660da8dR49

PR #23, opened 2023-01-18
Implement instance ids

shinypb:
We had instance ids back in the meshy-mesh prototype days, but they didn't make it into the agent until now. Instance ids are ephemeral UUIDs, unique to each session (that is, not persisted between runs).

Key changes:
- agent, queue, and storage instances all pick an instance id for themselves at startup
- advertising a service via mdns now requires you to provide an instance id, which will be exposed as the service's name (`mdns-sd` crate lingo) / instance name (macOS `dns-sd` command line tool lingo)
- proxied requests under the `/storage/` namespace will include a `Serval-Proxied-For` header (telling the recipient node the instance id of the node that's making the request) and will include `Serval-Proxied-From` headers in their responses back to the original caller telling which node the response ultimately came from

Other changes:
- Add `ServalError::AnyhowError` to make it easier to return any `anyhow::Error` from a function that returns a `ServalError`. This lets us use `?` in more places, and in fact, we should now be able to use it pretty much wherever just by throwing `.map_err(anyhow::Error)?` onto the end of an expression. Fingers crossed 🤞 that this aids in readability.
- Due to the above, I had to get rid of `ServalError::WasmError`, since thiserror's `derive_error` macro  is not able to handle the nuance of two different source error types that both happen to be built on top of `anyhow::Error`.


utils/src/errors.rs
```diff
@@ -28,9 +27,9 @@ pub enum ServalError {
     #[error("the WASM executable terminated abnormally; code={}", result.code)]
     AbnormalWasmExit { result: WasmResult },
 
-    /// The WASMtime engine responded with an error.
-    #[error("wasmtime engine error")]
-    WasmEngineError(#[from] wasi_common::Error),
+    // A conversion for anyhow::Error
+    #[error("an anyhow! error")]
```
shinypb, line 31: One thing I don't love about the way that this stuff works is that the original error gets lost—if you call `to_string` on an instance of `ServalError::AnyhowError`, you just end up with `"an anyhow! error"`, which is...less useful than I would've hoped.

agent/src/api/proxy.rs
```diff
@@ -44,13 +47,14 @@ pub async fn proxy_unavailable_services<B>(
 async fn proxy_request_to_service<B>(
     req: &Request<B>,
     service_name: &str,
+    source_instance_id: &Uuid,
 ) -> Result<Response, ServalError> {
     let node_info = discover_service(service_name).await.map_err(|err| {
         log::warn!("proxy_unavailable_services failed to find a node offering the service; service={service_name}; err={err:?}");
```
ceejbot, line 53: not in this pr, but noting that the log line mentions an outdated function name

agent/src/api/proxy.rs
```diff
@@ -79,19 +85,19 @@ async fn proxy_request_to_other_node<B>(
         }
         inner_req = inner_req.header(k, v);
     }
+
     inner_req = inner_req.header(
         "Serval-Proxied-For",
-        "<todo: put instance_id here once it exists in AppState>",
+        HeaderValue::from_str(&source_instance_id.to_string()).map_err(anyhow::Error::from)?,
```
ceejbot, line 91: Yeah, I feel your reaction to the clunk here, but hey. TODO done!

agent/src/api/proxy.rs
```diff
@@ -79,19 +85,19 @@ async fn proxy_request_to_other_node<B>(
...
     // Actually send the request
     let inner_req_res = inner_req.send().await?;
     let mut resp = reqwest_response_to_axum_response(inner_req_res).await?;
+
     resp.headers_mut().append(
         "Serval-Proxied-From",
-        "<todo: put instance_id from `info`'s properties here>"
-            .parse()
-            .unwrap(),
+        HeaderValue::from_str(&target_instance_id.to_string()).map_err(anyhow::Error::from)?,
```
ceejbot, line 100: ✅

utils/src/errors.rs
ceejbot: Grumpy noise. Let's live with it for now. Repeated theme: make forward progress and clean up later as we understand the problem space better.

utils/src/mdns.rs
```diff
@@ -11,6 +13,7 @@ use crate::networking::my_ipv4_addrs;
 pub fn advertise_service(
     service_name: &str,
     port: u16,
+    instance_id: &Uuid,
```
ceejbot, line 16: As I have read through this PR, the proliferation of arguments to these advertise/relay functions has bugged me, and it is at this point as I read that I understand how we might refactor them. Basic handwave: the runner state can hold not not just the instance id but also an idea of what services it offers, and manage the relaying. It can also cache recently-discovered services if we measure later and discover it's worth caching them. There's no need to do that refactor in this PR if you want to move on to new work. I continue to be happy to be a code Zamboni.

ceejbot (approved):
🦙

utils/src/mdns.rs
shinypb: Code Zamboni may be my new favourite term of art. I am very happy to benefit from your cleanup services, because it's good to see alternative approaches of things while I'm still in the learning zone.

agent/src/api/proxy.rs
shinypb: I was about to update this because you are right that the function name is wrong, but it's still the right name for the middleware itself, which is probably more correct from a reading-the-logs perspective.

PR #24, opened 2023-01-19
Host functions & better error messages

shinypb:
As discussed in Slack, this improves the derived ServalError variants to actually have useful error messages when written to log files or what-have-you.

Before: ```"an anyhow! error"```
After: ```"anyhow::Error: unknown import: `serval::add` has not been defined"```

This was done in the service of trying to figure out how to expose functions to the WASM jobs we run, which was trivial (of course) but surprisingly hard to figure out (I blame the lack of coffee). This PR adds a useless function to add two `i32`s together, which will have to do until we have a proper host SDK of some kind. The [call-a-host-function](https://github.com/servals/wasm-samples/tree/main/call-a-host-function) sample code makes use of this.

Despite the uselessness of their current form, both of these exist to be living documentation of how to expose things to jobs and how to import things from the job runner.

### DIFF
--- a/engine/src/lib.rs
+++ b/engine/src/lib.rs
@@ -31,6 +31,32 @@ impl ServalEngine {
         let mut linker: Linker<WasiCtx> = Linker::new(&engine);
         wasmtime_wasi::add_to_linker(&mut linker, |s| s)?;
 
+        // Wire up our host functions (e.g. functionality that we want to expose to the jobs we
+        // run). The first parameter to func_wrap is the name of the import namespace and the
+        // second is the name of the function. The default namespace for WASM imports is "env".
+        // For example, this:
+        // ```
+        // linker.func_wrap("env", "add", |a: i32, b: i32| -> i32 { a + b })?;
+        // ```
+        // will define a function at `env::add`, which you can access in your WASM job under the
+        // name "add" with the following extern block:
+        // ```
+        // extern "C" { fn add(a: i32, b: i32) -> i32; }
+        // ```
+        // If you'd like your function to be under a different namespace, define it like this...
+        // ```
+        // linker.func_wrap("foo", "add", |a: i32, b: i32| -> i32 { a + b })?;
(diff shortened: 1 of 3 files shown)


engine/src/lib.rs
```diff
@@ -31,6 +31,32 @@ impl ServalEngine {
...
+        // ```
+        // ...and import like this:
+        // ```
+        // #[link(wasm_import_module = "serval")]
+        // extern "C" { fn add(a: i32, b: i32) -> i32; }
+        // ```
+
+        // This exists solely so there's *something* for jobs to import, just so ensure the
+        // mechanisms all work.
+        linker.func_wrap("serval", "add", |a: i32, b: i32| -> i32 { a + b })?;
```
jkbecker, line 58: I don't understand the purpose of this line - wouldn't we want to wrap functions on a by-workload basis?

utils/src/errors.rs
```diff
@@ -28,7 +28,7 @@ pub enum ServalError {
     AbnormalWasmExit { result: WasmResult },
 
     // A conversion for anyhow::Error
-    #[error("an anyhow! error")]
+    #[error("anyhow::Error: {0}")]
```
jkbecker, line 31: I like how easy this fix ended up being

jkbecker (approved):
fine to merge, would just like to know what the func_wrap is doing in lib.rs

engine/src/lib.rs
shinypb: We might want to for some functions, and we might not want to for others. If the functionality is specific to a job (e.g. "what is my job id?") then maybe we'd want to set this up inside of the `execute` block (in which case we'd actually need a distinct `linker` instance, I think). But if the functionality we're exposing is universal (and, maybe, stateless), then we can set it up once and re-use it for all jobs.

jkbecker: got it, thanks

PR #25, opened 2023-01-19
Namespace all endpoints as /v1

ceejbot:
Did this in the cli, the endpoint uris, and the rust modules. Doing this any later in development will only get more annoying, so now is the time.

### DIFF
--- a/agent/src/structures/mod.rs
+++ b/agent/src/structures/mod.rs
@@ -1,0 +1,95 @@
+use anyhow::Result;
+use serde::{Deserialize, Serialize};
+use tokio::sync::Mutex;
+use utils::{blobs::BlobStore, errors::ServalError};
+use uuid::Uuid;
+
+use std::sync::Arc;
+use std::{collections::HashMap, path::PathBuf};
+
+/// Our application state. Fields are public for now but we'll want to fix that.
+#[derive(Debug, Clone, Serialize)]
+pub struct RunnerState {
+    pub instance_id: Uuid,
+    pub storage: Option<BlobStore>,
+    pub jobs: HashMap<String, JobMetadata>,
+    pub total: usize,
+    pub errors: usize,
+}
+
+impl RunnerState {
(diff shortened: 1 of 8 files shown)


agent/src/structures/mod.rs
```diff
@@ -0,0 +1,95 @@
...
+}
+
+impl JobMetadata {
+    pub fn new(name: String, description: String) -> Self {
+        let id = Uuid::new_v4();
+        Self {
+            id,
+            name,
+            description,
+            status_url: format!("/jobs/{}/status", id),
```
shinypb, line 55: Add /v1/ prefix here?

agent/src/structures/mod.rs
```diff
@@ -0,0 +1,95 @@
...
+            result_url: format!("/jobs/{}/result", id),
+        }
+    }
+
+    pub fn id(&self) -> &Uuid {
```
shinypb, line 60: Asking without judgement just because I am New Here: why add these getters?

agent/src/structures/mod.rs
ceejbot: Nice catch, ty!

ceejbot: This also reveals a structuring problem where the prefix is leaking out of the v1 module. I'll figure this out later.

shinypb: Agree; the URLs don't obviously need to be part of the JobMetadata struct, since they can be trivially generated given a job id.

agent/src/structures/mod.rs
ceejbot: Because the alternative was making the fields public, and while I did that for a different structure, I flinched at doing it here. This is not really a problem for us because we can move things between modules pretty freely right now, but at some point we're going to want to hide implementation details. And this one made me twitch.

shinypb (approved):

agent/src/structures/mod.rs
ceejbot: strong agree

PR #26, opened 2023-01-20
Wasmtime upgrade

shinypb:
This moves wasmtime and wasmtime-wasi up to the workspace-level Cargo.toml and bumps the version from 3.0.1 → 5.0.

I did this because I thought I needed something from it, but it turns out I do not. However, it's good to stay on the treadmill for core libraries like this.


ceejbot (approved):
That's one fast-moving project. 🐑

PR #29, opened 2023-01-26
agent - when using a randomly-selected port number, keep trying if binding fails

shinypb:
This should solve our "address in use" race condition when spinning up several agents in rapid succession.

### DIFF
--- a/agent/src/main.rs
+++ b/agent/src/main.rs
@@ -12,14 +12,14 @@ use axum::{
     extract::DefaultBodyLimit,
     middleware::{self},
     routing::{get, head, post, put},
-    Router,
+    Router, Server,
 };
 use dotenvy::dotenv;
 use tokio::sync::Mutex;
 use utils::{mdns::advertise_service, networking::find_nearest_port};
 use uuid::Uuid;
 
-use std::net::SocketAddr;
+use std::{net::SocketAddr, process};
 use std::{path::PathBuf, sync::Arc};
 
 mod api;
@@ -41,10 +41,6 @@ async fn main() -> Result<()> {
     env_logger::init();
 
     let host = std::env::var("HOST").unwrap_or_else(|_| "0.0.0.0".to_string());
-    let port: u16 = std::env::var("PORT")
-        .map(Ok)
-        .unwrap_or_else(|_| find_nearest_port(8100).map(|port| port.to_string()))?
-        .parse()?;
     let storage_role = match &std::env::var("STORAGE_ROLE").unwrap_or_else(|_| "auto".to_string())[..]
     {
         "always" => StorageRoleConfig::Always,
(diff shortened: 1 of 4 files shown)


agent/src/main.rs
```diff
@@ -92,20 +88,37 @@ async fn main() -> Result<()> {
         .layer(DefaultBodyLimit::max(MAX_BODY_SIZE_BYTES))
         .with_state(state);
 
-    let addr = format!("{}:{}", host, port);
-    log::info!("serval agent daemon listening on {}", &addr);
+    let predefined_port: Option<u16> = match std::env::var("PORT") {
```
ceejbot, line 91: I like this variable name.

shinypb, line 91: I spent quite a bit of time trying to write this as a map, but Rust was too picky about the error types being compatible.

````rust
let predefined_port: Option<u16> = std::env::var("PORT")
    .and_then(|port_str| port_str.parse::<u16>())
    .ok();
````

👇  

```
error[E0308]: mismatched types
  --> agent/src/main.rs:96:30
   |
96 |         .and_then(|port_str| port_str.parse::<u16>())
   |                              ^^^^^^^^^^^^^^^^^^^^^^^ expected enum `VarError`, found struct `ParseIntError`
   |
   = note: expected enum `Result<_, VarError>`
              found enum `Result<u16, ParseIntError>`

```

File: agent/src/main.rs [hunk 2]
```diff
@@ -92,20 +88,37 @@ async fn main() -> Result<()> {
...
+    };
+
+    // Start the Axum server; this is in a loop so we can try binding more than once in case our
+    // randomly-selected port number ends up conflicting with something else due to a race condition.
+    let mut port: u16;
+    let server: Server<_, _> = loop {
+        port = predefined_port.unwrap_or_else(|| find_nearest_port(8100).unwrap());
+        let addr: SocketAddr = format!("{host}:{port}").parse().unwrap();
+        match axum::Server::try_bind(&addr) {
+            Ok(builder) => break builder.serve(app.into_make_service()),
```
shinypb, line 103: I don't love how easy it is to lose the creation of the server, which is in here. It is returned as the result of the loop by using the `break` keyword (which, I must say, is a really nifty Rust feature).

agent/src/main.rs
```diff
@@ -92,20 +88,37 @@ async fn main() -> Result<()> {
...
+            Err(_) => {
+                // Port number in use already, presumably
+                if predefined_port.is_some() {
+                    log::error!("Specified port number ({port}) is already in use; aborting");
+                    process::exit(1);
+                }
+            }
+        }
+    };
```
ceejbot, line 112: I just tinkered with a couple different ways of writing this, trying to make the behavior if you have a passed-in port more clearly different, but this always ended up with more repeated code than what you have here. It's a tradeoff. I think this is readable enough!

ceejbot (approved):
🐫

PR #32, opened 2023-01-30
Switch from u32 → i64 for host/guest interop

### LINKED ISSUE #31: u32 → i64 for host/guest interop

shinypb:
Closes #31.

Not a ton of interesting stuff in here, and it's sort of hard to test because I don't have any 64-bit WASM jobs to play around with. However, this all appears to work.

### DIFF
--- a/engine/src/runtime/helpers.rs
+++ b/engine/src/runtime/helpers.rs
@@ -14,18 +14,24 @@ pub fn alloc<T>(
         return Err(anyhow!("Failed to get alloc function from guest"));
     };
 
-    // Note: We're casting an unsigned usize into an i32, which means we're losing
-    // half of our value range in wasm32. This shouldn't be a problem in practice,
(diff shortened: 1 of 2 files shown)


engine/src/runtime/helpers.rs
```diff
@@ -53,20 +59,20 @@ pub fn get_memory_from_caller<T>(caller: &mut Caller<'_, T>) -> Result<Memory, (
 pub fn read_bytes<T>(
     caller: &Caller<'_, T>,
     memory: Memory,
-    ptr: u32,
-    len: u32,
+    ptr: usize,
```
shinypb, line 62: Honestly, I am not sure how this will play out if we have a mix of 32- and 64-bit WASM payloads. 🤔 This usize will be based on the host, not the guest, and I am not sure how those two things will interact.

PR #34, opened 2023-01-31
SDK name changes and support for WASM extensions

shinypb:
The small news here is renaming “capabilities” to “extensions” per our [discussion in Slack](https://serval.slack.com/archives/C04BKH1J31S/p1675098898015979?thread_ts=1674764493.574109&cid=C04BKH1J31S). 🥱 

The big news here is support for WASM extensions. If you start the agent up with an `EXTENSIONS_PATH` environment variable set, it will enumerate all of the .wasm files in that directory and make them available as imports to all jobs that run on this node. Once we have the job queue implemented, nodes should send a list of all of their extensions to the job claim endpoint, manifests should include a list of extensions they require, and the job queue should only send jobs to nodes that have all of the required extensions.

What happens if a job requires something that doesn't exist? Well, for now, it just tries to run it anyway, which will fail, but I figure this is fine for now—perhaps some jobs will be smart enough to conditionally use imports if they're available, but degrade gracefully.

## Future considerations

- We can statically analyze WASM binaries to see which imports they're looking for, so the manifest thing might not end up being strictly required.
- We will eventually need to figure out a permission model for using extensions.

## Details
As an example, imagine this AssemblyScript file was compiled to `as-fib.wasm` extensions directory:

```ts
export function fib(n: i32): i32 {
  var a = 0, b = 1;
  if (n > 0) {
    while (--n) {
      let t = a + b;
      a = b;
      b = t;
    }
    return b;
  }
  return a;
}
```

Then, any job could import that function like this:

````rust
#[link(wasm_import_module = "as-fib")] // ← "as-fib" is the name of the .wasm file that should export this function
extern "C" {
    fn fib(n: i32) -> i32; // ← "fib" is the name of the exported function
}

fn main() {
    for i in 1..10 {
        println!("Fibonacci #{i} = {}", unsafe { fib(i) }); // ← calling into anything from extern requires an unsafe block
    }
}
````

## Try it yourself
The good news is that both of these files exist over in the wasm-samples repo — see [`as-fib.ts`](https://github.com/serval/wasm-samples/blob/main/assemblyscript/src/as-fib.ts) and [`sdk-test/src/main.rs`](https://github.com/serval/wasm-samples/blob/main/sdk-test/src/main.rs). If you check out a copy of that repo, you can try this for yourself.

First, build the assemblyscript artifacts:
````sh
just build assemblyscript
````

Then, start up the agent and point it at the wasm-samples repo's `build` directory

````sh
EXTENSIONS_PATH=../wasm-samples/build cargo run --bin serval-agent
````

You'll see something like this in the logs:
```
[2023-01-31T17:33:50Z INFO  serval_agent] Found 6 extensions at "../wasm-samples/build": ["loudify", "list-files", "as-hello", "as-fib", "sdk-test", "error-prone"]
```

Then, in another tab, run the sdk-test:

```sh
just build-and-run sdk-test
```

You'll see something like this:
```
Sending sdk-test.wasm (2,091,440 bytes for binary + payload) to serval agent...
----------
Calling a WASM extension...
Fibonacci #1 = 1
Fibonacci #2 = 1
Fibonacci #3 = 2
Fibonacci #4 = 3
Fibonacci #5 = 5
(...and so on)
```

### DIFF
--- a/agent/src/api/v1/jobs.rs
+++ b/agent/src/api/v1/jobs.rs
@@ -1,3 +1,5 @@
+use std::{collections::HashMap, path::PathBuf};
+
 use anyhow::Result;
 use axum::{
     extract::{Multipart, Path, State},
@@ -88,7 +90,7 @@ async fn run_job_inner(
         metadata.id(),
         binary.len()
     );
-    match execute_job(binary, input).await {
+    match execute_job(binary, input, &state.extensions).await {
         Ok(result) => {
             // We're not doing anything with stderr here.
             log::info!(
@@ -116,10 +118,14 @@ async fn run_job_inner(
 
 /// Run a job in the wasm engine.
 // Probably can vanish because there's only one caller.
-async fn execute_job(executable: Vec<u8>, input: Option<Vec<u8>>) -> Result<WasmResult> {
+async fn execute_job(
+    executable: Vec<u8>,
+    input: Option<Vec<u8>>,
+    extensions: &HashMap<String, PathBuf>,
+) -> Result<WasmResult> {
     let stdin = input.unwrap_or_default();
 
-    let mut engine = ServalEngine::new()?;
+    let mut engine = ServalEngine::new(extensions.clone())?;
     let result = engine.execute(&executable, &stdin)?;
 
     Ok(result)
--- a/agent/src/main.rs
+++ b/agent/src/main.rs
@@ -61,11 +61,13 @@ async fn main() -> Result<()> {
                 .unwrap_or_else(|_| std::env::temp_dir().join("serval_storage")),
         ),
     };
+    let extensions_path = std::env::var("EXTENSIONS_PATH").ok().map(PathBuf::from);
 
     let instance_id = Uuid::new_v4();
     let state = Arc::new(Mutex::new(RunnerState::new(
         instance_id,
         blob_path.clone(),
+        extensions_path.clone(),
     )?));
 
     const MAX_BODY_SIZE_BYTES: usize = 100 * 1024 * 1024;
@@ -86,7 +88,7 @@ async fn main() -> Result<()> {
         ))
         .route_layer(middleware::from_fn(clacks))
         .layer(DefaultBodyLimit::max(MAX_BODY_SIZE_BYTES))
-        .with_state(state);
+        .with_state(state.clone());
 
     let predefined_port: Option<u16> = match std::env::var("PORT") {
         Ok(port_str) => port_str.parse::<u16>().ok(),
@@ -117,6 +119,16 @@ async fn main() -> Result<()> {
     if blob_path.is_some() {
         log::info!("serval agent blob store mounted; path={blob_path:?}");
         advertise_service("serval_storage", port, &instance_id, None)?;
+    }
+    if let Some(extensions_path) = extensions_path {
+        let state = state.lock().await;
+        let extensions = &state.extensions;
+        log::info!(
+            "Found {} extensions at {extensions_path:?}: {:?}",
+            extensions.len(),
+            extensions.keys(),
+        );
+    } else {
     }
 
     server.await.unwrap();
--- a/agent/src/structures/mod.rs
+++ b/agent/src/structures/mod.rs
@@ -4,6 +4,7 @@ use tokio::sync::Mutex;
 use utils::{blobs::BlobStore, errors::ServalError};
 use uuid::Uuid;
 
+use std::fs;
 use std::sync::Arc;
 use std::{collections::HashMap, path::PathBuf};
 
@@ -11,6 +12,7 @@ use std::{collections::HashMap, path::PathBuf};
 #[derive(Debug, Clone, Serialize)]
 pub struct RunnerState {
     pub instance_id: Uuid,
+    pub extensions: HashMap<String, PathBuf>,
     pub storage: Option<BlobStore>,
     pub jobs: HashMap<String, JobMetadata>,
     pub total: usize,
@@ -18,15 +20,40 @@ pub struct RunnerState {
 }
 
 impl RunnerState {
-    pub fn new(instance_id: Uuid, blob_path: Option<PathBuf>) -> Result<Self, ServalError> {
+    pub fn new(
+        instance_id: Uuid,
+        blob_path: Option<PathBuf>,
+        extensions_path: Option<PathBuf>,
+    ) -> Result<Self, ServalError> {
         let storage = match blob_path {
             Some(path) => Some(BlobStore::new(path)?),
             None => None,
+        };
+        let extensions = if let Some(extensions_path) = extensions_path {
+            // Read the contents of the directory at the given path and build a HashMap that maps
+            // from the module's name (the filename minus the .wasm extension) to its path on disk.
+
+            let mut extensions: HashMap<String, PathBuf> = HashMap::new();
+            let dir_entries = fs::read_dir(&extensions_path)?;
+            for entry in dir_entries {
+                let Ok(entry) = entry else { continue };
+                let filename = entry.file_name();
+                let filename = filename.to_string_lossy();
+                if !filename.to_lowercase().ends_with(".wasm") {
+                    continue;
+                }
+                let module_name = &filename[0..filename.len() - ".wasm".len()];
+                extensions.insert(module_name.to_string(), entry.path());
+            }
+            extensions
+        } else {
+            HashMap::new()
         };
 
         Ok(RunnerState {
             instance_id,
             storage,
+            extensions,
             total: 0,
             errors: 0,
             jobs: HashMap::new(),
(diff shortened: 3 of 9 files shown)


agent/src/structures/mod.rs
```diff
@@ -4,29 +4,56 @@ use tokio::sync::Mutex;
...
+        instance_id: Uuid,
+        blob_path: Option<PathBuf>,
+        extensions_path: Option<PathBuf>,
+    ) -> Result<Self, ServalError> {
         let storage = match blob_path {
             Some(path) => Some(BlobStore::new(path)?),
             None => None,
         };
+        let extensions = if let Some(extensions_path) = extensions_path {
+            // Read the contents of the directory at the given path and build a HashMap that maps
```
shinypb, line 33: Since we enumerate the contents of the directory when the agent starts up, that means you can't hot-swap in new modules while the agent is running (although you can replace existing ones with new versions).

I've implemented it like this to avoid needing to do the work to enumerate available modules for every single job that runs. If the hot-swapping ends up being a problem, we can add a file watcher and/or a signal handler to force a re-scan.

engine/src/lib.rs
```diff
@@ -53,12 +63,39 @@ impl ServalEngine {
...
         let mut store = Store::new(&self.engine, wasi);
+
         let module = Module::from_binary(&self.engine, binary)?;
 
+        // Load any custom WASM node features that the job requires (...and that we have)
+        println!("Module wants these imports:");
+        let required_modules = module
+            .imports()
+            .map(|import| import.module().to_string())
+            .filter(|import| !import.starts_with("wasi_snapshot_"));
```
shinypb, line 74: Everything that uses WASI is going to try to import `wasi_snapshot_preview_1`; that's provided by `wasmtime_wasi` for us.

agent/src/api/v1/jobs.rs
```diff
@@ -88,7 +90,7 @@ async fn run_job_inner(
         metadata.id(),
         binary.len()
     );
-    match execute_job(binary, input).await {
+    match execute_job(binary, input, &state.extensions).await {
```
ceejbot, line 93: Noting this as something I'll need to revise when I integrate this work with my PR in progress. I reworked this specific flow extensively.

agent/src/api/v1/jobs.rs
```diff
@@ -116,10 +118,14 @@ async fn run_job_inner(
...
-async fn execute_job(executable: Vec<u8>, input: Option<Vec<u8>>) -> Result<WasmResult> {
+async fn execute_job(
+    executable: Vec<u8>,
+    input: Option<Vec<u8>>,
+    extensions: &HashMap<String, PathBuf>,
+) -> Result<WasmResult> {
     let stdin = input.unwrap_or_default();
 
-    let mut engine = ServalEngine::new()?;
+    let mut engine = ServalEngine::new(extensions.clone())?;
```
ceejbot, line 128: Noting this too.

agent/src/main.rs
```diff
@@ -61,11 +61,13 @@ async fn main() -> Result<()> {
                 .unwrap_or_else(|_| std::env::temp_dir().join("serval_storage")),
         ),
     };
+    let extensions_path = std::env::var("EXTENSIONS_PATH").ok().map(PathBuf::from);
 
     let instance_id = Uuid::new_v4();
     let state = Arc::new(Mutex::new(RunnerState::new(
         instance_id,
         blob_path.clone(),
+        extensions_path.clone(),
```
ceejbot, line 70: Checkmark. I have thoughts about RunnerState and exploding it, but they are as yet incomplete thoughts.

agent/src/main.rs
```diff
@@ -86,7 +88,7 @@ async fn main() -> Result<()> {
         ))
         .route_layer(middleware::from_fn(clacks))
         .layer(DefaultBodyLimit::max(MAX_BODY_SIZE_BYTES))
-        .with_state(state);
+        .with_state(state.clone());
```
ceejbot, line 91: Yeah. Eyebrow now all the way up about how I designed that state blob initially. No action required for this PR.

agent/src/main.rs
```diff
@@ -118,6 +120,16 @@ async fn main() -> Result<()> {
         log::info!("serval agent blob store mounted; path={blob_path:?}");
         advertise_service("serval_storage", port, &instance_id, None)?;
     }
+    if let Some(extensions_path) = extensions_path {
+        let state = state.lock().await;
+        let extensions = &state.extensions;
+        log::info!(
+            "Found {} extensions at {extensions_path:?}: {:?}",
+            extensions.len(),
+            extensions.keys(),
```
ceejbot, line 129: great log line

agent/src/structures/mod.rs
ceejbot: Agree! First implementation: simple; get more complicated when and if we have a reason.

agent/src/structures/mod.rs
```diff
@@ -4,29 +4,56 @@ use tokio::sync::Mutex;
...
+
+            let mut extensions: HashMap<String, PathBuf> = HashMap::new();
+            let dir_entries = fs::read_dir(&extensions_path)?;
+            for entry in dir_entries {
+                let Ok(entry) = entry else { continue };
+                let filename = entry.file_name();
+                let filename = filename.to_string_lossy();
+                if !filename.to_lowercase().ends_with(".wasm") {
+                    continue;
+                }
```
ceejbot, line 44: You can use [Path](https://doc.rust-lang.org/std/path/struct.Path.html) to break the full path into parent, stem, and extension if you prefer. The intent might be more clear, though tbh it is pretty clear this way too.

engine/src/lib.rs
ceejbot: I am tempted to put this comment right inline in the code.

engine/src/runtime/mod.rs
```diff
@@ -40,36 +38,36 @@ fn add(a: i32, b: i32) -> i32 {
     a + b
 }
 
-const INVOKE_CAPABILITY_ERROR_FAILED_TO_GET_MEMORY: i32 = -1;
-const INVOKE_CAPABILITY_ERROR_FAILED_TO_READ_CAPABILITY_NAME: i32 = -2;
-const INVOKE_CAPABILITY_ERROR_FAILED_TO_READ_DATA: i32 = -3;
-const INVOKE_CAPABILITY_ERROR_FAILED_TO_WRITE_RESPONSE: i32 = -4;
+const INVOKE_EXTENSION_ERROR_FAILED_TO_GET_MEMORY: i32 = -1;
```
ceejbot, line 41: Changing the jargon in variable names, check.

engine/src/runtime/mod.rs
```diff
@@ -40,36 +38,36 @@ fn add(a: i32, b: i32) -> i32 {
...
-    let capability_name = String::from_utf8_lossy(&buf);
+    let extension_name = String::from_utf8_lossy(&buf);
     let Ok(data) = read_bytes(&caller, memory, data_ptr, data_len) else {
         eprintln!("Failed to read from data_ptr");
-        return INVOKE_CAPABILITY_ERROR_FAILED_TO_READ_DATA;
+        return INVOKE_EXTENSION_ERROR_FAILED_TO_READ_DATA;
     };
 
-    let response = format!("Hello there! I can see that you tried to call the {capability_name} capability with {} bytes of data (to wit: {data:?}). Capabilities are not actually implemented yet, but this message did come from the host environment, so that's worth something, right?", data.len());
+    let response = format!("Hello there! I can see that you tried to call the {extension_name} extension with {} bytes of data (to wit: {data:?}). Extensions are not actually implemented yet, but this message did come from the host environment, so that's worth something, right?", data.len());
```
ceejbot, line 68: Love it.

ceejbot (approved):
One comment requested casually, but do 🐑  it

agent/src/structures/mod.rs
shinypb: Lovely, will do this. Thanks!

shinypb: Having actually done this, I think it ended up less readable with Path :( All of the Options and OsStrs end up blocking out the actual logic of the function:

````rust
for entry in dir_entries {
    let Ok(entry) = entry else { continue };
    let filename = entry.file_name();
    let filename = filename.to_string_lossy().to_string();
    let path = Path::new(&filename);
    let Some(ext) = path.extension() else { continue };
    if ext.to_ascii_lowercase() != "wasm" {
        continue;
    };
    let Some(module_name) = path.file_stem() else { continue };
    extensions.insert(module_name.to_string_lossy().to_string(), entry.path());
}
````

Compare that to the original:
````rust
for entry in dir_entries {
    let Ok(entry) = entry else { continue };
    let filename = entry.file_name();
    let filename = filename.to_string_lossy();
    if !filename.to_lowercase().ends_with(".wasm") {
        continue;
    }
    let module_name = &filename[0..filename.len() - ".wasm".len()];
    extensions.insert(module_name.to_string(), entry.path());
}
````

So, I'm going to leave it as-is for now, but I am delighted to learn about Path and look forward to using it elsewhere.

PR #35, opened 2023-01-31
Introduce ServalEngineError enum

shinypb:
CJ pointed out in [a comment on a prior PR](https://github.com/serval/serval-mesh/pull/27#pullrequestreview-1275808768) that we should probably start breaking `ServalError` up a bit more, and also stop just throwing anyhow:Errors around all willy-nilly. This PR is the first swing at that, introducing `ServalEngineError` as part of the engine crate.

I intentionally did not add an equivalent to `ServalError::AnyhowError` in the hopes of keeping the errors more focused, but this comes at the cost of having more error variants (currently 11).

One nice thing about this undertaking is that it's allowed me to delete this comment:
```
// This is a genuine error from the WASM engine, not a non-zero exit code from the
// the WASM executable. We report this as -1. Your improvements to this signaling
// method welcome.
```

This case is now represented as `ServalEngineError::ExecutionError`, which includes fields for stdout, stderr, and an inner `anyhow::Error`. Hopefully this is an actual improvement! 😅


engine/src/errors.rs
```diff
@@ -0,0 +1,42 @@
...
+    #[error("Reading or writing from guest memory failed")]
+    InteropMemoryAccessError(MemoryAccessError),
+
+    #[error("std::io::Error: {0}")]
+    IoError(#[from] std::io::Error),
+
+    #[error("Failed to load WASM module")]
+    ModuleLoadError(anyhow::Error),
+
+    #[error("Error reading bytes from stderr pipe")]
```
shinypb, line 37: One could argue that these could be collapsed into a single `StandardIOReadError`, since there's not really any actionable decision you could make in either case.

I am open to feedback on any of these error variants.

agent/src/api/v1/jobs.rs
```diff
@@ -103,12 +103,19 @@ async fn run_job_inner(
...
         }
+        Err(ServalEngineError::ExecutionError {
+            stdout: _,
+            error: _,
+            stderr,
+        }) => {
+            // Now the fun part of http error signaling: the request was successful, but the
+            // result of the operation was bad from the user's point of view. Our behavior here
+            // is yet to be defined but I'm sending back stderr just to show we can.
+            (StatusCode::OK, stderr).into_response()
```
ceejbot, line 117: ahahhahahahah
yes, this lets us do FAR better about error signaling

tbh I think the response structure can be better defined than we are doing it. needs a coordinated change with the cli.

agent/src/api/v1/jobs.rs
```diff
@@ -122,7 +129,7 @@ async fn execute_job(
     executable: Vec<u8>,
     input: Option<Vec<u8>>,
     extensions: &HashMap<String, PathBuf>,
-) -> Result<WasmResult> {
+) -> Result<WasmResult, ServalEngineError> {
```
ceejbot, line 132: nice

engine/src/errors.rs
ceejbot: I think this is a fantastic start, and being as informative as possible to the API consumer is best. Don't throw information away unless we have a good reason to throw it away.

engine/src/lib.rs
```diff
@@ -35,13 +37,16 @@ pub struct ServalEngine {
 
 impl ServalEngine {
     /// Create a new serval engine.
-    pub fn new(extensions: HashMap<String, PathBuf>) -> anyhow::Result<Self> {
+    pub fn new(extensions: HashMap<String, PathBuf>) -> Result<Self, ServalEngineError> {
         let engine = Engine::default();
         let mut linker: Linker<WasiCtx> = Linker::new(&engine);
-        wasmtime_wasi::add_to_linker(&mut linker, |s| s)?;
+        wasmtime_wasi::add_to_linker(&mut linker, |s| s)
+            .map_err(ServalEngineError::EngineInitializationError)?;
```
ceejbot, line 44: yeah this is worth the price of admission all the way <3

engine/src/runtime/helpers.rs
```diff
@@ -21,10 +22,13 @@ pub fn alloc<T>(
     let params: Vec<Val> = vec![Val::I32(num_bytes_required as i32)];
     let mut results: Vec<Val> = vec![Val::I32(0)];
 
-    alloc.call(caller, &params, &mut results)?;
+    if alloc.call(caller, &params, &mut results).is_err() {
+        return Err(ServalEngineError::InteropAllocFailed);
+    };
     let wasmtime::Val::I32(ptr) = results[0] else {
-        return Err(anyhow!("Alloc call failed"));
+        return Err(ServalEngineError::InteropAllocFailed);
```
ceejbot, line 29: lovely!

engine/src/runtime/helpers.rs
```diff
@@ -1,17 +1,18 @@
...
 /// does not expose a compatible alloc function, this will fail, and data exchange with the guest
 /// will not be possible. Our SDK automatically provides guest apps with this function.
 pub fn alloc<T>(
     caller: &mut Caller<'_, T>,
     num_bytes_required: usize,
-) -> Result<usize, anyhow::Error> {
+) -> Result<usize, ServalEngineError> {
     let Ok(alloc) = get_func_from_caller(caller, "alloc") else {
-        return Err(anyhow!("Failed to get alloc function from guest"));
+        return Err(ServalEngineError::InteropAllocUnavailable);
```
ceejbot, line 15: <3

engine/src/runtime/helpers.rs
```diff
@@ -79,6 +79,8 @@ pub fn write_bytes<T>(
     memory: &Memory,
     bytes: Vec<u8>,
 ) -> Result<usize, ServalEngineError> {
+    assert!(bytes.len() < u32::MAX as usize);
```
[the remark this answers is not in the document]
ceejbot: A nice use of assert.

ceejbot (approved):
This worked out great!

PR #39, opened 2023-02-02
Introduce the concept of a WASM manifest; use cacache for local storage

ceejbot:
This PR reworks the agent API and the cli to support a new concept of a "manifest". This is a toml specification of a WASM job type. It includes a name, a namespace, a path to a binary, and some human-readable descriptions.

Here's an example manifest:

```toml
name = "loudify"
namespace = "sh.serval"
binary = "../../wasm-samples/build/loudify.wasm"
version = "1"
description = "SHOUT SHOUT LET IT ALL OUT"
required_extensions = ["birdfeeder"]
```

Once a WASM manifest + executable has been stored in the system, you can execute that wasm using the
name in the manifest: `pounce run sh.serval.loudify inputfile.txt`

The local blob storage has been reimplemented to use the [cacache](https://lib.rs/crates/cacache)
crate. This is a content-adddressable store that *also* allows lookup of items by key. In our case
we store manifests with keys like `name.manifest.toml` and wasm executables with names like
`name.wasm`. Execution of jobs is now more indirect than it was, in that we look up the manifest by
name, and from that look up where the binary is by version. This indirection allows us to support
versioning more fully in the long term.

This PR also significantly reworks the agent API and changes the CLI to support this new API. The agent endpoints now look like this:

- `GET /monitor/ping`: cheap health check
- `POST /v1/jobs/:name/run`: body is interpreted as job input
- `GET /v1/storage/manifests`: fetch a list of manifests
- `POST /v1/storage/manifests`: store a new manifest
- `GET /v1/storage/manifests/:name`: get a manifest by name
- `HEAD /v1/storage/manifests/:name`: 404s if manifest not found
- `PUT /v1/storage/manifests/:name/executable/:version`: store a new wasm executable associated with the named task manifest
- `GET /v1/storage/manifests/:name/executable/:version`: get a wasm executable by manifest name & version
- `GET /v1/monitor/history`: very cheap run history; to be deleted
- `GET /v1/jobs`: list all running jobs; not implemented

Also, take a few minor updates to crates; revert change to the workspace syntax because it is making VSCode angry.

### DIFF
--- a/agent/src/api/v1/jobs.rs
+++ b/agent/src/api/v1/jobs.rs
@@ -33,14 +33,14 @@ pub async fn run_job(
         return (StatusCode::NOT_FOUND, "no manifest of that name found").into_response();
     };
 
-    let Ok(executable) = storage.executable_as_bytes(&name, &manifest.version()).await else {
+    let Ok(executable) = storage.executable_as_bytes(&name, manifest.version()).await else {
         return (StatusCode::NOT_FOUND, "no executable found for manifest; key={key}").into_response();
     };
 
     let job = Job::new(manifest, executable, input.to_vec());
     log::info!(
         "received WASM job; name={}; executable length={}; input length={}; id={}",
-        &job.manifest().name(),
+        &job.manifest().fq_name(),
         &job.executable().len(),
         input.len(),
         job.id()
--- a/agent/src/api/v1/jobs.rs
+++ b/agent/src/api/v1/jobs.rs
@@ -1,82 +1,55 @@
-use std::{collections::HashMap, path::PathBuf};
-
-use anyhow::Result;
 use axum::{
-    extract::{Multipart, Path, State},
+    body::Bytes,
+    extract::{Path, State},
     http::StatusCode,
-    response::{IntoResponse, Response},
+    response::IntoResponse,
     Json,
 };
 use engine::{errors::ServalEngineError, ServalEngine};
-use utils::structs::WasmResult;
+use utils::structs::Job;
 
 use crate::structures::*;
 
-/// This is the main worker endpoint. It accepts incoming jobs and runs them.
-pub async fn incoming(state: State<AppState>, mut multipart: Multipart) -> Response {
-    let mut envelope: Option<Envelope> = None;
-    let mut binary: Option<Vec<u8>> = None;
-    let mut input: Option<Vec<u8>> = None;
-
-    // chomp up form input here
-    while let Some(field) = multipart.next_field().await.unwrap() {
-        let name = field.name().unwrap().to_string();
-        let data = field.bytes().await.unwrap();
-        match name.as_str() {
-            "envelope" => {
-                let data = data.to_vec();
-                let Ok(parsed) = serde_json::from_slice(&data) else {
-                    // this is not good enough
-                    return (StatusCode::BAD_REQUEST, "job envelope is invalid".to_string()).into_response();
-                };
-                envelope = Some(parsed);
-            }
-            "input" => {
-                input = Some(data.to_vec());
-            }
-            "executable" => {
-                binary = Some(data.to_vec());
-            }
-            _ => {
-                log::info!("ignoring unknown field `{name}`");
-            }
-        }
-    }
-
-    let Some(binary) = binary else {
-        return (
-            StatusCode::BAD_REQUEST,
-            "no wasm executable data provided!".to_string(),
-        )
-            .into_response();
-    };
-
-    let envelope = envelope.unwrap_or_default();
-    let metadata: JobMetadata = JobMetadata::from(envelope);
-    log::info!(
-        "received WASM job; name={}; executable length={}; input length={}",
-        metadata.name(),
-        binary.len(),
-        input.as_ref().map(|input| input.len()).unwrap_or_else(|| 0),
-    );
-
-    run_job_inner(state, metadata, binary, input).await
+/// Report on runtime history
+pub async fn monitor_status(state: State<AppState>) -> Json<RunnerState> {
+    let state = state.lock().await;
+    Json(state.clone())
 }
 
-async fn run_job_inner(
+/// Get running jobs
+pub async fn running(_state: State<AppState>) -> impl IntoResponse {
+    StatusCode::NOT_IMPLEMENTED
+}
+
+/// This is the main worker endpoint. It accepts incoming jobs and runs them.
+pub async fn run_job(
+    Path(name): Path<String>,
     state: State<AppState>,
-    metadata: JobMetadata,
-    binary: Vec<u8>,
-    input: Option<Vec<u8>>,
-) -> Response {
-    let mut state = state.lock().await;
+    input: Bytes,
+) -> impl IntoResponse {
+    let mut lock = state.lock().await;
+    let storage = lock.storage.as_ref().unwrap();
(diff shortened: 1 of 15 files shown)


agent/src/api/v1/jobs.rs
```diff
@@ -1,82 +1,55 @@
...
+
+/// This is the main worker endpoint. It accepts incoming jobs and runs them.
+pub async fn run_job(
+    Path(name): Path<String>,
+    state: State<AppState>,
+    input: Bytes,
+) -> impl IntoResponse {
+    let mut lock = state.lock().await;
+    let storage = lock.storage.as_ref().unwrap();
+    let Ok(manifest) = storage.manifest(&name).await else {
```
shinypb, line 32: Does this mean that only nodes that have storage locally available can run jobs?

utils/src/errors.rs
```diff
@@ -64,12 +86,13 @@ impl IntoResponse for ServalError {
...
                 StatusCode::BAD_REQUEST
             }
-            ServalError::AnyhowError(_) => StatusCode::INTERNAL_SERVER_ERROR,
             ServalError::BlobAddressInvalid(_) => StatusCode::BAD_REQUEST,
             ServalError::BlobAddressNotFound(_) => StatusCode::NOT_FOUND,
             ServalError::IoError(_) => StatusCode::NOT_FOUND,
             ServalError::ServiceNotFound => StatusCode::NOT_FOUND,
-            ServalError::ReqwestError(_) => StatusCode::INTERNAL_SERVER_ERROR,
+            ServalError::AnyhowError(_) => StatusCode::INTERNAL_SERVER_ERROR,
+            // Catch-all for anything we don't want to add specific status codes for.
```
shinypb, line 94: Want to nuke the ones above that already map to this?

utils/src/structs.rs
```diff
@@ -8,3 +15,129 @@ pub struct WasmResult {
...
+pub struct Job {
+    /// The ID of this job.
+    id: Uuid,
+    /// Fully-qualified job manifest specification. E.g., "sh.serval.birdfeeder"
+    manifest: Manifest,
+    /// bytes for the wasm executable
+    executable: Vec<u8>,
+    /// Input data
+    input: Vec<u8>,
+    // TODO: might have version chosen to run here, plus run options; might also store the input
```
shinypb, line 103: Right, something like an `Option<u16>` to reference a specific version number (rather "latest") would make sense.

agent/src/api/v1/jobs.rs
ceejbot: At the moment, yes. I think the immediate next followup work is to make a node ask if if has the appropriate binary, and if not, find it.

cli/src/main.rs
```diff
@@ -86,55 +87,56 @@ fn build_url(path: String, version: Option<&str>) -> String {
 }
 
 fn upload_manifest(manifest_path: PathBuf) -> Result<()> {
-    println!("Reading manifest at {}", manifest_path.display());
+    println!("Reading manifest: {}", manifest_path.display());
```
shinypb, line 90: Is this is the equivalent of
```rust
println!("Reading manifest: {manifest_path:?}");
```
?

utils/src/structs.rs
```diff
@@ -57,17 +45,18 @@ impl Manifest {
...
-        self.version.clone()
+    pub fn version(&self) -> &str {
+        &self.version
     }
 
     /// Get the fully-qualified-by-namespace name for this job type manifest.
-    pub fn name(&self) -> String {
-        self.fq_name.clone()
+    pub fn fq_name(&self) -> String {
+        let name = self.name.to_ascii_lowercase().replace('-', "_");
```
shinypb, line 58: Is the `-` support just because of our wasm-samples naming convention?

agent/src/api/v1/jobs.rs
shinypb: Makes sense. Thoughts on swapping `BlobStore` out for an abstract trait that has `CacacheBlobStore` as one implementation and `RemoteBlobStore` (something that proxies to a serval_storage instance) as the other? If that sounds reasonable to me I am happy to start chipping away at it.

cli/src/main.rs
ceejbot: Not a debugging print, but converting a pointer to a pathbuf to its display version. I think it's cleaner; haven't looked at what the output is with the :? formatter.

utils/src/structs.rs
ceejbot: This should get a TODO next to it, because IMO it needs to be cleaned more definitively to restrict it to allowed characters.

cli/src/main.rs
```diff
@@ -41,44 +42,102 @@ struct Args {
...
+        format!("{baseurl}/{path}")
+    }
+}
+
+fn upload_manifest(manifest_path: PathBuf) -> Result<()> {
+    println!("Reading manifest: {}", manifest_path.display());
+    let manifest = Manifest::from_file(&manifest_path)?;
+
+    println!("Reading WASM executable:{}", manifest.binary().display());
+    let executable = read_file(manifest.binary().to_path_buf())?;
```
shinypb, line 94: Can we make this resolve the binary path relative to manifest_path? I'd like to have my manifest live at wasm-samples/serval-facts/manifest.toml and look like this:

```toml
name = "serval-facts"
namespace = "sh.serval"
binary = "../build/serval-facts.wasm"
# ...snip
```

Right now, that won't work because the code tries to load the binary relative to the current working directory, not relative to the manifest.

cli/src/main.rs
```diff
@@ -41,44 +42,102 @@ struct Args {
...
+        .build()?;
+
+    // Start building pretty output now that we're past the most likely errors.
+    println!();
+    let mut table = Table::new();
+    table.set_format(*prettytable::format::consts::FORMAT_CLEAN);
+    table.add_row(row!["WASM task name:", manifest.fq_name()]);
+    table.add_row(row!["Version:", manifest.version()]);
+
+    let url = build_url("storage/manifests".to_string(), Some("1"));
```
shinypb, line 107: We pass `Some("1")` around often enough that a `build_url_v1` helper function probably wouldn't be a bad idea at some point.

cli/src/main.rs
```diff
@@ -41,44 +42,102 @@ struct Args {
...
+        manifest.version()
+    );
+    let url = build_url(vstring, Some("1"));
+    let response = client.put(url).body(executable).send()?;
+    if response.status().is_success() {
+        let wasm_integrity = response.text()?;
+        table.add_row(row!["WASM integrity:", wasm_integrity]);
+        table.add_row(row![
+            "To run:",
+            format!("cargo run -- run {}", manifest.fq_name()).bold().blue()
```
shinypb, line 132: Consider making this `cargo run -p serval`, which will work correctly whether you're in the `cli` subdirectory or not.

shinypb:
This looks good to me, modulo the rustfmt error giving you grief. 😅 Calling out that landing this means that `STORAGE_ROLE=never` nodes can no longer run jobs, which (temporarily) means that we can't have nodes on more than one machine working together due to our lack of distributed storage.

cli/src/main.rs
ceejbot: Agreed! It annoyed me.

PR #40, opened 2023-02-03
Fix request proxying, introducing RUNNER_ROLE

shinypb:
First and most embarrassingly of all, this fixes `proxy_unavailable_services` to actually include the body of the outermost request when proxying to another node. I am not sure how I noticed that this wasn't working. 😮‍💨

Secondly, this introduces the ability to proxy requests in the `/v1/jobs/` namespace to nodes advertising the `_serval_runner` role.

Finally, this adds a `RUNNER_ROLE` environment variable that controls whether a given instance should run jobs or proxy them elsewhere. The default is "auto", which enables job running if and only if the current platform is supported by wasmtime's cranelift code generator. This is in support of an observation that @jkbecker made about his Raspberry Pi Zero, which is a 32-bit ARM machine and is not supported by cranelift.

### DIFF
--- a/agent/src/api/v1/proxy.rs
+++ b/agent/src/api/v1/proxy.rs
@@ -1,5 +1,6 @@
 use anyhow::Result;
 use axum::{
+    body::{Body, HttpBody},
     extract::State,
     http::{Request, StatusCode},
     middleware::Next,
@@ -15,12 +16,13 @@ use utils::{
 use super::*;
 use crate::structures::AppState;
 
-pub async fn proxy_unavailable_services<B>(
+pub async fn proxy_unavailable_services(
     State(state): State<AppState>,
-    req: Request<B>,
-    next: Next<B>,
+    mut req: Request<Body>,
+    next: Next<Body>,
 ) -> Result<Response, StatusCode> {
     let path = req.uri().path();
+
     if path.starts_with("/v1/storage/") {
         let state = state.lock().await;
         if state.storage.is_none() {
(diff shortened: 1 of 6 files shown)


agent/src/api/v1/proxy.rs
```diff
@@ -1,52 +1,66 @@
...
-
-            let Ok(resp) = proxy_request_to_service(&req, "_serval_storage", &state.instance_id).await else {
-                // Welp, not much we can do
-                return Ok((StatusCode::SERVICE_UNAVAILABLE, "Storage not available").into_response());
-            };
-
-            return Ok(resp);
-        }
+
+    let runner_state = state.lock().await;
```
shinypb, line 27: The downside of this refactor is that the middleware now briefly gets a lock on the state for all requests. The upside is that the code for `get_proxy_target_service` is much simpler and more readable.

Feedback welcome.

ceejbot: I am going to tear into this state thing this afternoon. We can probably make the locks cover smaller sets of shared data.

An additional thought I had is that we can structure this at startup time, when the routes are mounted. Check config; register one of two flavors of function at the mount point, either the do-it-locally function or the proxy function. This approach would make each endpoint more readable and probably let us write some tidy proxying code as well. 

We could do both things and tighten all of this up nicely, I think.

engine/src/lib.rs
```diff
@@ -165,6 +166,13 @@ impl ServalEngine {
 
         Ok(result)
     }
+
+    pub fn is_available() -> bool {
+        // The cranelift code generator that underpins wasmtime doesn't support every architecture
+        // under the sun; in particular, it doesn't support 32-bit ARM, which is a potentially
+        // viable target for the Serval agent in general.
+        Isa::from_arch(std::env::consts::ARCH).is_some()
```
ceejbot, line 174: ✅

ceejbot (commented):
Note comment about a way we might clean this up. What do you think?

shinypb:
Are you imagining the routes looking something like this?

```rust
.route("/v1/storage/manifests", post(if storage_role { v1::storage::store_manifest) } else { proxy })
```

or this?
```rust
let app = Router::new()
    .route("/monitor/ping", get(ping))
    .route("/monitor/status", get(v1::jobs::monitor_status));

if storage_role {
    app.route("/v1/storage/manifests", get(v1::storage::list_manifests))
        route("/v1/storage/manifests", post(v1::storage::store_manifest));
} else {
    app.route("/v1/storage/manifests", get(proxy_to("_serval_storage")))
        route("/v1/storage/manifests", post(proxy_to("_serval_storage")));
}
```

or something different entirely?

PR #41, opened 2023-02-04
Moved the storage global into a once_cell

ceejbot:
And removed the hacky runtime state tracking. The axum app state object is now much slimmer, and we can lose the mutex around it.

The cell we're sticking the blob store into is a leverage point where we can add some indirection later if we want to.

### DIFF
--- a/agent/src/api/v1/jobs.rs
+++ b/agent/src/api/v1/jobs.rs
@@ -3,7 +3,6 @@ use axum::{
     extract::{Path, State},
     http::StatusCode,
     response::IntoResponse,
-    Json,
 };
 use engine::{errors::ServalEngineError, ServalEngine};
 use utils::structs::Job;
@@ -11,9 +10,8 @@ use utils::structs::Job;
 use crate::structures::*;
 
 /// Report on runtime history
-pub async fn monitor_status(state: State<AppState>) -> Json<RunnerState> {
-    let state = state.lock().await;
-    Json(state.clone())
+pub async fn monitor_status(_state: State<AppState>) -> impl IntoResponse {
+    StatusCode::NOT_IMPLEMENTED
 }
 
 /// Get running jobs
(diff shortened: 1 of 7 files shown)


agent/src/api/v1/jobs.rs
```diff
@@ -3,17 +3,15 @@ use axum::{
...
 use engine::{errors::ServalEngineError, ServalEngine};
 use utils::structs::Job;
 
 use crate::structures::*;
 
 /// Report on runtime history
-pub async fn monitor_status(state: State<AppState>) -> Json<RunnerState> {
-    let state = state.lock().await;
-    Json(state.clone())
+pub async fn monitor_status(_state: State<AppState>) -> impl IntoResponse {
```
shinypb, line 13: This endpoint was giving me some grief on Friday, too

agent/src/api/v1/proxy.rs
```diff
@@ -24,20 +25,19 @@ pub async fn proxy_unavailable_services(
...
 
-        let Ok(resp) = proxy_request_to_service(&mut req, &target_service, &runner_state.instance_id).await else {
+        let Ok(resp) = proxy_request_to_service(&mut req, &target_service, &state.instance_id).await else {
             // Welp, not much we can do
             return Ok((StatusCode::SERVICE_UNAVAILABLE, format!("{target_service} not available")).into_response());
         };
         return Ok(resp);
     }
-    drop(runner_state); // must explicitly drop before we call `next.run(req)` lest we deadlock
+    drop(state); // must explicitly drop before we call `next.run(req)` lest we deadlock
```
shinypb, line 40: I think we might not need this drop anymore, since we're no longer locking

shinypb:
Nice simplification. :+1:

Side note: I find the API for once_cell's set kind of surprising, since it returns `Err(the_value_you_tried_to_set)` if it's already been set. I wonder why they did that. 🤔

shinypb (approved):

agent/src/api/v1/proxy.rs
ceejbot: Fact.

PR #42, opened 2023-02-17
Refactor the proxy mount routing extensively

ceejbot:
Each`v1` api submodule now exports `mount()` and `mount_proxy()` as the functions to call to mount all their routes. This corrals all knowledge of exact endpoint paths to the modules, so the details of the path are no longer something `main()` knows about. Rewrote the middleware to be handlers mounted with the `any` verb. There's now a little code duplication in each of the proxying endpoints, but this does not yet hit my threshold for generalizing it. I do like that this approach let me hide all of the endpoint handlers inside their modules.

Additionally, removed all dependencies on native-tls from our usage of reqwest. In theory, this will make us more easily portable. In theory.

### DIFF
--- a/agent/src/api/v1/storage.rs
+++ b/agent/src/api/v1/storage.rs
@@ -2,9 +2,8 @@ use axum::{
     body::{Body, Bytes, StreamBody},
     extract::{Path, State},
     http::{header, Request, StatusCode},
-    middleware::Next,
-    response::{Response, IntoResponse},
-    routing::{get, head, post, put},
+    response::IntoResponse,
+    routing::{any, get, head, post, put},
     Json,
 };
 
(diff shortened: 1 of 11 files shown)


agent/src/api/v1/storage.rs
```diff
@@ -11,8 +13,53 @@ use utils::structs::Manifest;
...
+        )
+}
+
+/// Relay all storage requests to a node that can handle them.
+pub async fn proxy(
+    State(state): State<AppState>,
+    mut req: Request<Body>,
+    next: Next<Body>,
+) -> Result<Response, StatusCode> {
+    // TODO: I kinda don't like this as middleware, but maybe it's the cleanest way to do it.
```
shinypb: Are there any other alternatives that are any better than putting something at the top of each individual handler? We could write that so it wouldn't be particularly hideous, but it's still boilerplate:

```rust
async fn get_executable(
     Path((name, version)): Path<(String, String)>,
     State(_state): State<AppState>,
 ) -> impl IntoResponse {

    if let Some(proxy_req) = maybe_proxy_request(&mut req, &state) {
        let resp = proxy_req.await;
        return Ok(resp);
    }

    // ... rest of handler goes here
````

File: agent/src/main.rs [hunk 2]
```diff
@@ -97,36 +98,25 @@ async fn main() -> Result<()> {
...
-        )
-        // end optional endpoints
-        .route_layer(middleware::from_fn_with_state(
-            state.clone(),
-            v1::proxy::proxy_unavailable_services,
-        ))
+        .route("/monitor/status", get(monitor_status));
+
+    // NOTE: We have two of these now. If we develop a third, generalize this pattern.
+    router = if state.has_storage {
```
shinypb, line 107: Genuinely asking: would it be better to pass state into `v1::storage::mount` so it can be the one to install the middleware instead?

shinypb:
I love this refactor; it's a big improvement. Nice work! :sparkles:

shinypb (approved):

agent/src/main.rs
ceejbot: I think this is a reasonable suggestion! I'll do another refactoring pass.

PR #43, opened 2023-03-09
wapm.io integration

jkbecker:
Since I was planning on looking into executing arbitrary existing software packages via Serval Mesh, I had the sudden urge of making [wapm.io](https://wapm.io/) integration happen. My goal for this change is to make it as easy as 

```bash
❯ cargo run --bin serval -- pull author/package/1.0.0
Found package wapm.io/author/package/1.0.0
Downloading binary https://registry-cdn.wapm.io/contents/author/package/1.0.0/target/wasm32-wasi/release/package.wasm... Done

 WASM task name:      io.wasm.author.package
 Version:             1.0
 Manifest integrity:  sha256-p7XKW+adaldkmlakmsdlmalsdkml=
 WASM integrity:      sha256-a68Q3adlasldfkmalsdfkaslkdfjlsE8uYwVTafrg=
 To run:              cargo run -p serval -- run io.wasm.author.package
```

Where `serval pull` would

1. pull the appropriate package from wapm.io
2. generate a manifest for it
3. run `serval store` on the manifest.

First actual code submission, please be gentle 😸 

This is a very rough work in progress - just wanted to share it already because my Rust is, well, rusty, and you might have conceptual opinions as well...

Please let me know what you think!


cli/src/main.rs
```diff
@@ -286,7 +292,19 @@ fn blocking_maybe_discover_service_url(
...
+
+fn pull(registry: &str, identifer: String) -> Result<()> {
+    let package_spec = RegistryPackageSpec::parse(registry, identifer).unwrap();
+    println!("{:#?}", package_spec);
+    let dl_url = package_spec.registry.baseurl_download.to_string();
+    println!("{dl_url}");
+
+    println!("{}", package_spec.fqtn());
+
+    println!("{}", package_spec.fqdn());
```
ceejbot, line 306: This function doesn't exist on the type; is this is a copy-paste slip-up?

cli/src/main.rs
```diff
@@ -286,7 +292,19 @@ fn blocking_maybe_discover_service_url(
     };
 
     let port = info.get_port();
-    return Ok(format!("http://{addr}:{port}"));
+    Ok(format!("http://{addr}:{port}"))
+}
+
+fn pull(registry: &str, identifer: String) -> Result<()> {
+    let package_spec = RegistryPackageSpec::parse(registry, identifer).unwrap();
+    println!("{:#?}", package_spec);
```
ceejbot, line 300: I love this approach to sketching out implementations, because it is literally what I do too. Print intermediate results, then when they're right, remove the logging and use the data.

cli/src/main.rs
```diff
@@ -320,6 +338,7 @@ fn main() -> Result<()> {
         Command::Status { id } => status(id)?,
         Command::History => history()?,
         Command::Ping => ping()?,
+        Command::Pull { identifer } => pull("wapm", identifer)?,
```
ceejbot, line 341: I have a surprisingly long reaction to this single line, which is in my overall PR comment!

utils/src/registry.rs
```diff
@@ -0,0 +1,101 @@
...
+}
+
+impl RegistryPackageSpec {
+
+    /// Parses a registry package using a registry key and a package identifier
+    pub fn parse(registry: &str, identifer: String) -> Result<RegistryPackageSpec, ServalError> {
+        let registry_spec = Registry::from_str(registry).unwrap();
+        // Split the string by "/". We expect the provided identifier to be one of the following:
+        // - author/package/version
+        // - author/package@version
```
ceejbot, line 45: I suggest that we pick one and standardize on it. I further suggest that we go with `<author>/<package>[@version]` with the version spec being optional. No version: use latest.

ceejbot (commented):
I love this. This is a great idea. We need this.

Let's think about feature design a little bit!

I know of two WASM package manager thingies:

- https://warg.io
- https://wapm.io

WAPM is real. WARG is a proposal with a reference implementation that's under active development that might become real. If we generalize, we should support both registries. This leads us to a cli invocation like this:

```shell
serval pull wapm author/package@version
serval pull wapm ceejbot/loudbot
serval pull warg bytecodealliance.org/dog-facts@1.0.0
```

Note that I'm going with `@version` optionally because that's what all the other package manager clis seem to do.

We can pass the registry string into the `pull()` function in the cli implementation, and then it can be parsed by the specific registry implementation.

I have some specific suggestions about how we handle the specific registry implementations. Right now you're passing a string into a `parse()` function, which goes on to fill out some struct fields with registry-specific format strings at runtime. I suggest we restructure to make Rust's type system do this work for us.

Here's a possible implementation:

```rust
enum PackageRegistry {
    Wapm,
    Warg
}

impl PackageRegistry {
    fn download_url(&self, pkg: &PackageSpec) -> String {
        match self {
            PackageRegistry::Wapm => {
                format!("https://registry-cdn.wapm.io/contents/{}/{}/{}/target/wasm32-wasi/release/{}.wasm", pkg.author, pkg.name, pkg.version, pkg.name)
            },
            PackageRegistry::Warg => todo!(),
        }
    }

    // even cooler....
    fn download(&self, pkg: &PackageSpec) -> Result<Bytes, ServalError> {
        // do the work of downloading from this kind of registry
    }
}

pub struct PackageSpec {
    author: String,
    name: String,
    version: String,
    registry: PackageRegistry,
    // any other useful fields
}

impl PackageSpec {
    // other useful functions here

    pub fn download_url(&self) -> String {
        self.registry.download_url(self)
    }
}

impl TryFrom<&str> for PackageSpec {
    // put your parsing code here
}
```

The advantage of this approach is that different registries might have different ways to download, e.g., different ways to authenticate. We can hide that inside a specific registry `download()` implementation if we need to and get rid of the `download_url()` function. You can probably think of a bunch of different ways to achieve the goal of hiding the registry-specific stuff as much as possible.

The interesting Rust lesson here is to notice that I swapped out your match on a string approach for matching on an enum.

utils/src/registry.rs
```diff
@@ -0,0 +1,101 @@
+use std::str::FromStr;
```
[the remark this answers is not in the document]
jkbecker: Will do that! thx

jkbecker:
Thanks, these are awesome suggestions. I'm so not used to having an "implementation" for an Enum - I initially had a similar setup to yours with a Registry enum (but different / less idiomatic implementation downstream from it) but then discarded it because I got stuck somewhere. Definitely rewriting this based on your suggestion! :)

cli/src/main.rs
```diff
@@ -289,6 +296,88 @@ fn blocking_maybe_discover_service_url(
...
+                    println!("🛑 Server error: {}", status_code);
+                    println!("   There may be an issue with this package manager.");
+                } else if status_code.is_client_error() {
+                    println!("🛑 Client error: {}", status_code);
+                    println!("{:#?}", status_code);
+                    if status_code == 404 {
+                        println!("   Failed to download from {:?}", pkg_spec.download_urls());
+                    }
+                    println!();
+                    if pkg_spec.version == "latest" && pkg_spec.registry == PackageRegistry::Wapm {
```
jkbecker, line 340: I am not happy about the lengthy special case handling and hint printing inline here. It makes this function super long and hard to read. 
What's a good way of outsourcing these hints to somewhere else? Just a helper function in `registry.rs` which matches on a keyword and prints these hints? something better?
Note: these are lazy temporary hints that could get automated away over time.

utils/src/registry.rs
```diff
@@ -0,0 +1,306 @@
...
+        self.binary_path().exists()
+    }
+}
+
+/// Converts an identifier string to a `PackageSpec`
+impl TryFrom<std::string::String> for PackageSpec {
+    type Error = ServalError;
+    /**
+    This function matches a package specification string.
+    It supports a number of variants:
```
jkbecker, line 135: I may be missing some variants... but at least they are all automatically tested ;)

utils/src/registry.rs
```diff
@@ -0,0 +1,322 @@
...
+                vec![
+                    // For some very stupid reason, wasm binaries can sit in multiple locations. Hopefully this is the full list:
+                    format!("https://registry-cdn.wapm.io/contents/{}/{}/{}/{}.wasm", pkg.author, pkg.name, pkg.version, pkg.module),
+                    format!("https://registry-cdn.wapm.io/contents/{}/{}/{}/target/wasm32-wasi/release/{}.wasm", pkg.author, pkg.name, pkg.version, pkg.module)
+                ]
+            }
+            PackageRegistry::Warg => todo!(),
+        }
+    }
+    // even cooler....
```
ceejbot, line 89: hey wait I recognize that ellipsis with an extra dot

ceejbot (commented):
As before, some optional comments aimed at being rusty!

cli/src/main.rs
```diff
@@ -289,6 +295,100 @@ fn blocking_maybe_discover_service_url(
     Ok(format!("http://{addr}:{port}"))
 }
 
+/// Pull a Wasm package from a package manager, generate its manifest, and store it.
+fn pull(identifer: String) -> Result<()> {
+    let pkg_spec = PackageSpec::try_from(identifer).unwrap();
```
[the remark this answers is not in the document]
jkbecker: ahhhh... nice. I was theoretically aware of let else but did not grok the specific usage. Will hunt down a bunch of `unwrap()`s and clean them up.

utils/src/registry.rs
jkbecker: blatant plagiarism. painstakingly embellished for hours and hours, though, I claim fair use 😹

shinypb (commented):
Posting a comment solely to get GitHub to stop showing me this PR under my Review Requests :joy: Hope you're well!

PR #44, opened 2023-03-14
Start emitting metrics.

ceejbot:
This commit adds the `metrics` crate, which is a commonly-used set of macros for emitting metrics used by many output generators, similar to the `log` and `tracing` crates. Typical metrics look like this:

```rust
metrics::increment_counter!("process:start");
metrics::histogram!("memory", someNumberHere);
metrics::increment_gauge!("gauge_name", someNumberHere);
```

This is literally all you need to do to send a new metric. The crate offers more affordances if you need them or want to re-use a specific metric emitter in a function.

Added basic metrics to each of the endpoints in the agent, as well as some process restart metrics.

Added one output sink, which we'll want to replace immediately because it's only suitable as a hello-world proof that metrics are being sent. In order to see metrics from the process, you'll need to check out my fork of the metrics-rs repo and run the `metrics-observer` subproject.

```terminal
gh repo clone ceejbot/metrics-rs-fork
cd metrics-rs-fork/metrics-observer
cargo run -- 0.0.0.0:9000
```

Then run the agent and do some things. This is obviously unacceptable long-term, but we'll pick a first real candidate for metrics output eventually. In the short term, it's good to get into the habit of instrumenting code as we write it.

### DIFF
--- a/agent/src/api/v1/jobs.rs
+++ b/agent/src/api/v1/jobs.rs
@@ -26,6 +26,7 @@ pub fn mount_proxy(router: ServalRouter) -> ServalRouter {
 async fn proxy(State(state): State<AppState>, mut request: Request<Body>) -> impl IntoResponse {
     let path = request.uri().path();
     log::info!("relaying a job runner request; path={path}");
+    metrics::increment_counter!("proxy:{path}");
 
     if let Ok(resp) =
         super::proxy::relay_request(&mut request, SERVAL_SERVICE_RUNNER, &state.instance_id).await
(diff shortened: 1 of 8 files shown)


agent/src/api/v1/jobs.rs
```diff
@@ -26,13 +26,15 @@ pub fn mount_proxy(router: ServalRouter) -> ServalRouter {
...
     log::info!("relaying a job runner request; path={path}");
+    metrics::increment_counter!("proxy:{path}");
 
     if let Ok(resp) =
         super::proxy::relay_request(&mut request, SERVAL_SERVICE_RUNNER, &state.instance_id).await
     {
         resp
     } else {
         // Welp, not much we can do
+        metrics::increment_counter!("proxy:{path}:error");
```
shinypb, line 37: Thoughts on changing this to `proxy:error:` so wildcard searches will be easier to write?

shinypb (approved):
Super cool!

PR #46, opened 2023-03-22
Job permissions

shinypb:
## What
This implements permission checks for jobs. I wanted to represent these in a nice format in the manifest files, so I ended up having to write my own serializing/deserializing code, which ended up taking up a lot more time than I expected. Alas. I think the format is pleasant:

- `proc:read:*` — whether they are allowed to read `/proc`; presumably a `proc:write:*` will make sense someday. Implementing more granular paths is an exercise for the future; my current thinking is that we'll have e.g. `proc:read:foo` to give access to `/proc/foo`, but maybe that's weird and we'll end up with a separate mechanism for giving access to various filesystem paths. Who knows!
- `extension:foo` — permission to use the extension `foo`
- `extension:*` — permission to use any extension
- `http:example.com` — permission to make HTTP requests to example.com
- 'http:*' — permission to make HTTP requests to any domain (unclear whether this should even exist as a feature, but it's useful for now)

These permissions can be placed in manifest files under `required_permissions`, and can be overridden when the test-runner via the `--permissions foo,bar,baz` parameter.

I also spruced up the extension loading code a bit, creating a `ServalExtension` struct to handle some of the file system shenanigans.

### DIFF
--- a/agent/src/api/v1/jobs.rs
+++ b/agent/src/api/v1/jobs.rs
@@ -66,8 +66,8 @@ async fn run_job(
     let job = Job::new(manifest, executable, input.to_vec());
     log::info!(
         "received WASM job; name={}; executable length={}; input length={}; id={}",
-        &job.manifest().fq_name(),
-        &job.executable().len(),
+        job.manifest().fq_name(),
+        job.executable().len(),
         input.len(),
         job.id()
     );
@@ -77,9 +77,9 @@ async fn run_job(
     // What we'll do later is accept this job for processing and send it to a thread or something.
     // But for now we do it right here, in our handler.
     // The correct response by design is a 202 Accepted plus the metadata object.
-    // TODO: SER-38 - capture exit code for failed jobs
     log::info!(
-        "about to run job name=TODO; id={}; executable size={}",
+        "about to run job name={}; id={}; executable size={}",
+        job.manifest().fq_name(),
         job.id(),
         job.executable().len()
     );
--- a/agent/src/api/v1/jobs.rs
+++ b/agent/src/api/v1/jobs.rs
@@ -89,7 +89,14 @@ async fn run_job(
     let Ok(mut engine) = ServalEngine::new(extensions) else {
         return (StatusCode::INTERNAL_SERVER_ERROR, "unable to create wasm engine").into_response();
     };
-    let result = engine.execute(job.executable(), job.input());
+
+    // todo: verify that the user who submitted the job is actually authorized for all of the
+    // permissions that are listed in the manifest. If not, return a 403 error.
+    let result = engine.execute(
+        job.executable(),
+        job.input(),
+        job.manifest().required_permissions(),
+    );
 
     match result {
         Ok(result) => {
--- a/engine/src/lib.rs
+++ b/engine/src/lib.rs
@@ -94,8 +94,8 @@ impl ServalEngine {
             .filter(|import| !import.starts_with("wasi_snapshot_"))
             // Our SDK functions are exported under the serval namespace; this is set up in the
             // register_exports function that we call in our constructor, above.
-            .filter(|import| import != "serval");
-        let required_modules: HashSet<String> = HashSet::from_iter(required_modules);
+            .filter(|import| import != "serval")
+            .collect::<HashSet<String>>();
 
         log::info!("Job wants the following extensions: {required_modules:?}");
 
--- a/engine/src/lib.rs
+++ b/engine/src/lib.rs
@@ -7,18 +7,22 @@
     unused_qualifications
 )]
 
-use std::collections::{HashMap, HashSet};
+use std::{
+    collections::{HashMap, HashSet},
+    fs::File,
+    path::PathBuf,
+};
 
 use anyhow::anyhow;
 use cranelift_codegen_meta::isa::Isa;
 use extensions::ServalExtension;
-use utils::structs::WasmResult;
+use utils::structs::{Permission, WasmResult};
 use wasi_common::{
     pipe::{ReadPipe, WritePipe},
     I32Exit,
 };
 use wasmtime::{Engine, Linker, Module, Store};
-use wasmtime_wasi::{WasiCtx, WasiCtxBuilder};
+use wasmtime_wasi::{Dir, WasiCtx, WasiCtxBuilder};
 
 pub mod errors;
 pub mod extensions;
@@ -43,15 +47,6 @@ impl ServalEngine {
         wasmtime_wasi::add_to_linker(&mut linker, |s| s)
             .map_err(ServalEngineError::EngineInitializationError)?;
 
-        // link the experimental HTTP support
-        let http_state = HttpState::new().map_err(ServalEngineError::EngineInitializationError)?;
-        http_state
-            .add_to_linker(&mut linker, |_| HttpCtx {
-                allowed_hosts: Some(vec!["insecure:allow-all".to_string()]),
-                max_concurrent_requests: Some(42),
-            })
-            .map_err(ServalEngineError::EngineInitializationError)?;
-
         // Wire up our host functions (functionality that we want to expose to the jobs we run)
         register_exports(&mut linker).map_err(|_| {
             ServalEngineError::EngineInitializationError(anyhow!("Failed to register exports"))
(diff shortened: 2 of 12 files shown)


agent/src/api/v1/jobs.rs
```diff
@@ -77,9 +77,9 @@ async fn run_job(
     // What we'll do later is accept this job for processing and send it to a thread or something.
     // But for now we do it right here, in our handler.
     // The correct response by design is a 202 Accepted plus the metadata object.
-    // TODO: SER-38 - capture exit code for failed jobs
     log::info!(
-        "about to run job name=TODO; id={}; executable size={}",
+        "about to run job name={}; id={}; executable size={}",
```
jkbecker, line 81: hah, yesterday I looked at some logs and was like "I should find out why this job is called TODO" :D

engine/src/lib.rs
```diff
@@ -69,20 +62,66 @@ impl ServalEngine {
...
         input: &[u8],
+        // List of elevated permissions for this binary
+        permissions: &[Permission],
     ) -> Result<WasmResult, ServalEngineError> {
         let stdout = WritePipe::new_in_memory();
         let stderr = WritePipe::new_in_memory();
 
+        // Link the experimental HTTP support
+        let allowed_http_hosts = if permissions.contains(&Permission::AllHttpHosts) {
+            // todo: unclear whether we should actually support a wildcard like this
```
jkbecker, line 77: yeah I'd like to slow-walk this wildcard idea a little bit. i can see applications that would need it but let's make sure the restricted case is rock-solid first

engine/src/lib.rs
```diff
@@ -69,20 +62,66 @@ impl ServalEngine {
...
+        };
+
+        if !allowed_http_hosts.is_empty() {
+            let http_state =
+                HttpState::new().map_err(ServalEngineError::EngineInitializationError)?;
+            http_state
+                .add_to_linker(&mut self.linker, move |_| HttpCtx {
+                    // todo: there's some confusion around whether
+                    allowed_hosts: Some(allowed_http_hosts.clone()),
+                    max_concurrent_requests: Some(42),
```
jkbecker, line 96: _Can_ a single Wasm module actually make 42 different concurrent requests, considering it single-threaded?

engine/src/lib.rs
```diff
@@ -69,20 +62,66 @@ impl ServalEngine {
...
-            .build();
+            .stderr(Box::new(stderr.clone()));
+
+        // Give the engine access to whichever parts of the file system are required
+        // TODO: this list should be pulled from the job's manifest, and permissions should be
+        // checked against the owner of the job in question and the configuration of this node (that
+        // is, the list of file system locations that any job by any user can access should be
+        // defined at the node level, and then the subset of those that any job by a specific user
+        // can access should exist in our configuration store, and then the subset of those that a
+        // specific job by that specific user should exist in the manifest for that job. Phew!
```
jkbecker, line 113: Yes!

engine/src/lib.rs
```diff
@@ -69,20 +62,66 @@ impl ServalEngine {
...
+        if permissions.contains(&Permission::ProcRead) {
+            let path = PathBuf::from("/proc");
```
jkbecker, line 115: Would it make sense to test for `std::env::consts::OS == "linux"` (or a better test for whether we're running on linux) before even attempting this, in order to return a more helpful error? This could be confusing if proc access is explicitly pointed out as a feature and then returns "unsupported feature". It sure is unsupported, but for OS reasons specific to the host it is running on, not because the overall runtime doesn't support it yet.

test-runner/src/main.rs
```diff
@@ -6,46 +6,90 @@
...
+    {
+        "toml" => {
+            let manifest = Manifest::from_file(&args.exec_path)?;
+
+            eprintln!(
+                "{} {}",
+                "manifest:".blue().bold(),
+                toml::to_string(&manifest)
+                    .unwrap()
+                    .replace('\n', "\n          ")
```
jkbecker, line 62: Man, I stared at this line for a good while until it clicked. Not a critique, I definitely need this in another place 😅

utils/src/structs.rs
```diff
@@ -30,25 +30,63 @@ pub struct Manifest {
...
+            description: String::from(""),
+            required_extensions: vec![],
+            required_permissions: vec![],
+        }
+    }
+
     pub fn from_string(input: &str) -> Result<Self, ServalError> {
         let manifest: Manifest = toml::from_str(input)?;
+        if manifest.binary.is_relative() {
+            return Err(ServalError::RelativeBinaryPathInManifestError);
```
jkbecker, line 60: The CLI currently assumes that a Wasm path is _always_ relative (and in fact, a file in the same directory as the manifest file itself). [see here](https://github.com/serval/serval-mesh/blob/main/cli/src/main.rs#L94)
Absolute paths don't really make sense in a manifest considering that both the manifest and the binary end up in the mesh storage anyways, rendering the path obsolete.
Actually... shouldn't there simply be a convention for where the Wasm binary should be relative to the manifest? I feel like the way the CLI handles it is too much of an explicit manipulation as-is, but it shouldn't really be absolute, either.

jkbecker: ping: @ceejbot probably has opinions here

jkbecker: Ah, crap. A file obviously can't be relative to a string, I get why this must throw an error.
Still... Do we then ultimately store the manifest in the mesh with an absolute path that is invalid on all systems except the originating one? 🤔

utils/src/structs.rs
```diff
@@ -30,25 +30,63 @@ pub struct Manifest {
...
+        }
         Ok(manifest)
     }
 
     pub fn from_file(path: &PathBuf) -> Result<Self, ServalError> {
         let buf = std::fs::read_to_string(path)?;
-        let manifest: Manifest = toml::from_str(&buf)?;
+        let mut manifest = toml::from_str::<Manifest>(&buf)?;
+        if manifest.binary.is_relative() {
+            // If the binary file actually exists, replace its relative path with absolute path. (If
```
jkbecker, line 69: see commentary above on relative vs absolute paths

utils/src/structs.rs
```diff
@@ -130,3 +168,71 @@ impl Job {
...
+    HttpHost(String),
+}
+
+impl Display for Permission {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        let str = match self {
+            Permission::ProcRead => String::from("proc:read"),
+            Permission::AllExtensions => String::from("extension:*"),
+            Permission::Extension(name) => format!("extension:{name}"),
+            Permission::AllHttpHosts => String::from("http:*"),
```
jkbecker, line 187: Reading `http:` somewhere makes me go "but where is `https:`"?
Does this actually just support HTTP, or does it actually support arbitrary TCP/UDP traffic? In which case, how about `web:` or `connection:` or ...?

engine/src/lib.rs
shinypb: That's my thinking as well, although it's preferable to have it while we're in wild hacking mode.

engine/src/lib.rs
shinypb: Yeah, because the actual work is being done over in Rust Land.

engine/src/lib.rs
shinypb: Fun fact: no. 😂 At first, I implemented this using `#[cfg(target_os = "linux")]` and it ended up leading to (incorrect) dead code warnings whenever I compiled it on macOS. 🤦

utils/src/structs.rs
shinypb: The manifests we store in the mesh shouldn't have binary paths at all; they should have content-addressable storage hashes. Binary paths only make sense when running locally (e.g. when you're in the act of storing a job, or when you're using the test-runner).

utils/src/structs.rs
shinypb: It should support HTTP requests of any kind, including unencrypted, TLS, http/2, http/3, you name it. Try not to read `http:` as the first part of a URL and try to read it as a namespace for the permissions instead.

jkbecker: I'm certainly trying but my stupid brain always goes "where's the slash slash after the colon" 😂

utils/src/structs.rs
```diff
@@ -130,3 +168,71 @@ impl Job {
...
+#[derive(Clone, Debug, Eq, Hash, PartialEq)]
+pub enum Permission {
+    ProcRead,
+    AllExtensions,
+    Extension(String),
+    AllHttpHosts,
+    HttpHost(String),
+}
+
+impl Display for Permission {
```
ceejbot, line 181: Yeah, writing out predictable code like this feels bad, but here we are.

utils/src/errors.rs
```diff
@@ -70,6 +70,9 @@ pub enum ServalError {
     /// Translation for errors from ssri.
     #[error("ssri::Error: {0}")]
     SsriError(#[from] ssri::Error),
+
+    #[error("Manifest with a relative binary path was passed to Manifest::from_string; only absolute paths are supported here")]
+    RelativeBinaryPathInManifestError,
```
ceejbot, line 75: I like this specific limitation quite a bit, btw.

test-runner/src/main.rs
```diff
@@ -58,27 +102,59 @@ fn main() -> anyhow::Result<()> {
...
     };
 
+    let extensions = args
+        .extensions_path
+        .map(|extensions_path| {
+            load_extensions(&extensions_path).expect("Failed to load extensions")
+        })
+        .unwrap_or_default();
+    {
+        let extensions_readable = if extensions.is_empty() {
```
ceejbot, line 112: There are two places in this small use case where list-of-extensions would benefit from being a little generalization: in this implementation of display, and in the override-or-manifest-list usage below. I don't think it rises to the level of doing something yet, but I note this out loud because I had this reaction and I'd have shared that if we'd been pairing on this.

ETA: There are actually three places. See the validating-perm-strings code above. (Yes, I'm reading it backwards right now.)

test-runner/src/main.rs
```diff
@@ -6,46 +6,90 @@
...
+        _ => Manifest::new(&args.exec_path),
+    };
+
+    let permissions_override = args.permissions.map(|raw_perms| {
+        raw_perms
+            .split(',')
+            .map(|raw_perm| {
+                Permission::from_str(raw_perm).unwrap_or_else(|_| {
+                    eprintln!("Error: Invalid permission '{raw_perm}'");
+                    exit(1);
```
ceejbot, line 76: Safest response.

test-runner/src/main.rs
```diff
@@ -6,46 +6,90 @@
...
+        .to_string_lossy()
+        .as_ref()
+    {
+        "toml" => {
+            let manifest = Manifest::from_file(&args.exec_path)?;
+
+            eprintln!(
+                "{} {}",
+                "manifest:".blue().bold(),
+                toml::to_string(&manifest)
```
ceejbot, line 60: Enable the `display` crate feature and use `to_string_pretty()` here!

ceejbot (approved):
Minor suggestions, mostly about opportunities for abstractions that help the cli.

engine/src/lib.rs
```diff
@@ -69,20 +62,66 @@ impl ServalEngine {
     /// Run the passed-in WASM executable on the given input bytes.
     pub fn execute(
         &mut self,
+        // WebAssembly binary to execute
         binary: &[u8],
+        // Data to pass to WebAssembly as stdin
         input: &[u8],
+        // List of elevated permissions for this binary
```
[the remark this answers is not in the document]
shinypb: Good call; I will do a pass

test-runner/Cargo.toml
```diff
@@ -10,5 +10,9 @@ atty.workspace = true
 clap.workspace = true
 engine = { path = "../engine" }
 owo-colors = "3.5.0"
+serde.workspace = true
+serde_json.workspace = true
+toml = "0.7.3"
```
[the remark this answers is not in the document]
shinypb: Will fix, thank you

test-runner/src/main.rs
shinypb: In JavaScript Land I would've written something like `"manifest: ".replace(/./g, " ")` but it would be so much more verbose in  Rust Land...

test-runner/src/main.rs
shinypb: These ended up looking identical:
````
          namespace = ""
          version = "1"
          binary = "/Users/mark/Code/serval/wasm-samples/build/serval-facts.wasm"
          description = "fun facts"
          required_extensions = []
          required_permissions = []
````

````
manifest: name = "serval-facts"
          namespace = ""
          version = "1"
          binary = "/Users/mark/Code/serval/wasm-samples/build/serval-facts.wasm"
          description = "fun facts"
          required_extensions = []
          required_permissions = []
````

test-runner/src/main.rs
shinypb: Good call. Do you have any thoughts on a particularly elegant solution?

utils/src/errors.rs
shinypb: Me too, although I have a follow-up PR that softens it by letting you pass `base_path: Option<PathBuf>` into `Manifest::from_string`, which I did as part of a refactor to let `from_string` and `from_file` share the same code path (`::from_file` now reads the file and passes it into `::from_string`).

PR #48, opened 2023-03-24
Switch from mdns to using the kaboodle peer mesh

ceejbot:
The agent and the cli now both attempt to join a network of peers, established via kaboodle, and use that network instead of mdns to discover relevant services. 

The most interesting implementation details are in the file `utils/src/mesh.rs`. This file includes some wrapper traits that probably belong in Kaboodle itself, since they were designed to be re-usable and to encode the responsibilities of any Kaboodle implementation: KaboodleMesh and KaboodlePeer. There are implementations plus some serval-specific functions in the structs ServalMesh and PeerMetadata. (All names are negotiable. I already don't like my initial choices.)

The peer metadata struct gets a little complicated because I attempted to provide some future-proofing via a way to version it. It's encoded via `binencode`, with a version envelope wrapper to allow us to make newer agents that change what gets packed into identity able to interoperate with older agents, and let older agents ignore newer ones. I welcome improvements to this somewhat ad-hoc design.

Agent roles are now an enum, `ServalRoles`. The `strum` crate provides Display and FromString implementations. This adds some safety and validity-checking to what had been some magic strings.

We knew we'd stop blocking in the cli some day, and this is that day: the cli is now fully async.

Finally, and most noisily, corrected the spelling of "Wasm" everywhere. Apologies for that.

### DIFF
--- a/agent/src/api/v1/proxy.rs
+++ b/agent/src/api/v1/proxy.rs
@@ -7,12 +7,13 @@ use http::{
     header::{CONTENT_LENGTH, EXPECT, HOST},
     HeaderValue,
 };
-use mdns_sd::ServiceInfo;
 use utils::{
     errors::ServalError,
-    mdns::{discover_service, get_service_instance_id},
+    mesh::{KaboodlePeer, PeerMetadata},
 };
 use uuid::Uuid;
+
+use crate::structures::MESH;
 
 // Relay the given request to to the first node that we discover that is advertising the given
 // service. in the future, we may keep a list of known nodes for a given service so we can avoid
@@ -22,15 +23,16 @@ pub async fn relay_request(
     service_name: &str,
     source_instance_id: &Uuid,
 ) -> Result<Response, ServalError> {
-    let node_info = discover_service(service_name).await.map_err(|err| {
-        log::warn!("proxy_unavailable_services failed to find a node offering the service; service={service_name}; err={err:?}");
+    let mesh = MESH.get().expect("Peer network not initialized!");
+    let Some(peer) = mesh.find_role(&service_name.to_string()).await else {
+        log::warn!("proxy_unavailable_services failed to find a node offering the service; service={service_name}");
         metrics::increment_counter!("proxy:no_service");
-        err
-    })?;
+        return Err(ServalError::ServiceNotFound);
+    };
 
-    let result = proxy_request_to_other_node(req, &node_info, source_instance_id).await;
+    let result = proxy_request_to_other_node(req, &peer, source_instance_id).await;
     result.map_err(|err| {
-        log::warn!("Failed to proxy request to other node; node={node_info:?}; err={err:?}");
+        log::warn!("Failed to proxy request to peer; peer={peer:?}; err={err:?}");
         metrics::increment_counter!("proxy:failure");
         err
     })
@@ -38,19 +40,23 @@ pub async fn relay_request(
 
 async fn proxy_request_to_other_node(
     req: &mut Request<Body>,
-    info: &ServiceInfo,
+    peer: &PeerMetadata,
     source_instance_id: &Uuid,
 ) -> Result<Response, ServalError> {
-    let target_instance_id = get_service_instance_id(info)?;
-    let host = info.get_addresses().iter().next().unwrap(); // unwrap is safe because discover_service will never return a service without addresses
-    let port = info.get_port();
+    let target_instance_id = peer.name();
+    let Some(socket_addr) = peer.address() else {
+        log::warn!("got a peer without an address; peer={}", peer.name());
+        metrics::increment_counter!("proxy:failure");
+        return Err(ServalError::ServiceNotFound);
+    };
+
     let path = req.uri().path();
     let query = req
         .uri()
         .query()
         .map(|qs| format!("?{qs}"))
         .unwrap_or_else(|| "".to_string());
-    let url = format!("http://{host}:{port}{path}{query}");
+    let url = format!("http://{socket_addr}{path}{query}");
     let mut inner_req = reqwest::Client::new().request(req.method().clone(), url);
 
     // Copy over the headers, modulo a few that are only relevant to the original request
@@ -86,7 +92,7 @@ async fn proxy_request_to_other_node(
 
     resp.headers_mut().append(
         "Serval-Proxied-From",
-        HeaderValue::from_str(&target_instance_id.to_string()).map_err(anyhow::Error::from)?,
+        HeaderValue::from_str(target_instance_id).map_err(anyhow::Error::from)?,
     );
 
     Ok(resp)
--- a/agent/src/api/v1/proxy.rs
+++ b/agent/src/api/v1/proxy.rs
@@ -9,7 +9,7 @@ use http::{
 };
 use utils::{
     errors::ServalError,
-    mesh::{KaboodlePeer, PeerMetadata, ServalRole},
+    mesh::{PeerMetadata, ServalRole},
 };
 use uuid::Uuid;
 
@@ -24,13 +24,15 @@ pub async fn relay_request(
     source_instance_id: &Uuid,
 ) -> Result<Response, ServalError> {
     let mesh = MESH.get().expect("Peer network not initialized!");
-    let Some(peer) = mesh.find_role(role).await else {
+
+    let candidates = mesh.find_role(role).await;
+    let Some(peer) = candidates.first() else {
         log::warn!("proxy_unavailable_services failed to find a node offering the service; service={role}");
         metrics::increment_counter!("proxy:no_service");
         return Err(ServalError::ServiceNotFound);
     };
 
-    let result = proxy_request_to_other_node(req, &peer, source_instance_id).await;
+    let result = proxy_request_to_other_node(req, peer, source_instance_id).await;
     result.map_err(|err| {
         log::warn!("Failed to proxy request to peer; peer={peer:?}; err={err:?}");
         metrics::increment_counter!("proxy:failure");
@@ -44,11 +46,7 @@ async fn proxy_request_to_other_node(
     source_instance_id: &Uuid,
 ) -> Result<Response, ServalError> {
     let target_instance_id = peer.instance_id();
-    let Some(socket_addr) = peer.address() else {
-        log::warn!("got a peer without an address; peer={}", peer.instance_id());
-        metrics::increment_counter!("proxy:failure");
-        return Err(ServalError::ServiceNotFound);
-    };
+    let http_address = peer.http_address();
 
     let path = req.uri().path();
     let query = req
@@ -56,7 +54,7 @@ async fn proxy_request_to_other_node(
         .query()
         .map(|qs| format!("?{qs}"))
         .unwrap_or_else(|| "".to_string());
-    let url = format!("http://{socket_addr}{path}{query}");
+    let url = format!("{http_address}{path}{query}");
     let mut inner_req = reqwest::Client::new().request(req.method().clone(), url);
 
     // Copy over the headers, modulo a few that are only relevant to the original request
--- a/agent/src/api/v1/proxy.rs
+++ b/agent/src/api/v1/proxy.rs
@@ -9,7 +9,7 @@ use http::{
 };
 use utils::{
     errors::ServalError,
-    mesh::{KaboodlePeer, PeerMetadata},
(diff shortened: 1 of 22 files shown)


utils/src/mesh.rs
```diff
@@ -0,0 +1,113 @@
...
+    fn identity(&self) -> Vec<u8>;
+    fn name(&self) -> &str;
+    fn roles(&self) -> Vec<String>;
+}
+
+/// A little wrapper around kaboodle so we can hide the machinery of encoding and decoding.
+/// the identity payload.
+#[async_trait]
+pub trait KaboodleMesh {
+    type A;
```
shinypb: Aren't we all 🤣

utils/src/mesh.rs
```diff
@@ -0,0 +1,113 @@
...
+    async fn new(
+        name: &str,
+        roles: Vec<String>,
+        port: u16,
+        interface: Option<Interface>,
+    ) -> Result<Box<Self>, KaboodleError>;
+    /// Get a list of peers. It's the implementer's responsibility to decide if this is fresh or cached somehow.
+    async fn peers(&self) -> Vec<Self::A>;
+    /// Given a specific role, look for a peer that advertises the role.
+    async fn find_role(&self, role: &str) -> Result<Self::A, KaboodleError>;
```
shinypb: Perhaps this will change in a subsequent commit that I haven't looked at yet, but should this be `find_roles` (plural) and return every known peer that has a given role? Because, for example, if we are e.g. looking up a file stored on the mesh, we'll need the instance ids of every storage node to figure out who to ask.

agent/src/main.rs
```diff
@@ -171,14 +168,20 @@ async fn main() -> Result<()> {
         );
     }
 
+    let mut roles: Vec<String> = Vec::new();
+    if blob_path.is_some() {
```
shinypb, line 174: Doesn't really matter, but if you change this to `if let Some(blob_path) = blob_path {` then you can get rid of the `:?` in the format string and the output will look nicer.

utils/src/mesh.rs
```diff
@@ -5,109 +5,136 @@ use kaboodle::{errors::KaboodleError, Kaboodle};
...
+        self.inner.roles.clone()
+    }
 }
 
-impl KaboodlePeer for ServalPeer {
-    fn new(address: SocketAddr, encoded: Vec<u8>) -> Self {
-        // TODO: this is actually fallible
+impl KaboodlePeer for PeerMetadata {
+    fn from_identity(address: SocketAddr, encoded: Vec<u8>) -> Self {
+        // TODO: this is actually fallible; when might it fail?
```
shinypb, line 102: This data is coming in over unreliable transport, so it's possible for a bit to get flipped and lead to a corrupted encoded payload.

agent/src/api/v1/proxy.rs
```diff
@@ -22,35 +23,40 @@ pub async fn relay_request(
     service_name: &str,
     source_instance_id: &Uuid,
 ) -> Result<Response, ServalError> {
-    let node_info = discover_service(service_name).await.map_err(|err| {
-        log::warn!("proxy_unavailable_services failed to find a node offering the service; service={service_name}; err={err:?}");
+    let mesh = MESH.get().expect("Peer network not initialized!");
```
shinypb, line 26: (Thinking aloud 🤔 ) Is it worth promoting this to `ServalError`, or (since this happening would be a Programming Error™) is the catastrophic nature of the `.expect` preferable?

agent/src/api/v1/proxy.rs
```diff
@@ -22,35 +23,40 @@ pub async fn relay_request(
...
     req: &mut Request<Body>,
-    info: &ServiceInfo,
+    peer: &PeerMetadata,
     source_instance_id: &Uuid,
 ) -> Result<Response, ServalError> {
-    let target_instance_id = get_service_instance_id(info)?;
-    let host = info.get_addresses().iter().next().unwrap(); // unwrap is safe because discover_service will never return a service without addresses
-    let port = info.get_port();
+    let target_instance_id = peer.name();
+    let Some(socket_addr) = peer.address() else {
```
shinypb: I am making this comment in the wrong place, but why does `peer.address() return an `Option`?

agent/src/main.rs
```diff
@@ -42,15 +42,15 @@ async fn main() -> Result<()> {
     env_logger::init();
 
     // TODO: metrics sink initialization based on env vars or config
-    let addr: SocketAddr = "0.0.0.0:9000".parse().unwrap();
+    let metrics_port = std::env::var("METRICS_PORT").unwrap_or_else(|_| "9000".to_string());
```
shinypb, line 45: This reminds me about something Johannes brought up the other day; all of our port bindings in serval-mesh are for IPv4 (`0.0.0.0:{port}`) rather than IPv6 (`::{port}`, I believe, although I have not tried it).

Created https://github.com/serval/serval-mesh/issues/50 to track it.

agent/src/main.rs
```diff
@@ -139,6 +139,8 @@ async fn main() -> Result<()> {
         Err(_) => None,
     };
 
+    let host = std::env::var("HOST").unwrap_or_else(|_| "0.0.0.0".to_string());
```
shinypb, line 142: Now that we have `MESH_PORT`, Should we rename `HOST` and `PORT` to `HTTP_HOST` and `HTTP_PORT`?

cli/src/main.rs
```diff
@@ -262,35 +266,46 @@ fn ping() -> Result<()> {
...
-    let Some(addr) = info.get_addresses().iter().next() else {
-        // this should not ever happen, but computers
-        return Err(anyhow!(format!(
-            "Discovered a node that has no addresses",
-        )));
+    let metadata = PeerMetadata::new(format!("client@{host}"), vec![ServalRole::Client], None);
+    let mut mesh = ServalMesh::new(metadata, mesh_port, None).await?;
+    mesh.start().await?;
+
+    // There has to be a better way.
```
shinypb, line 297: I shall work on this way today ✨

cli/src/main.rs
```diff
@@ -262,35 +266,46 @@ fn ping() -> Result<()> {
     Ok(())
 }
 
-fn blocking_maybe_discover_service_url(
-    service_type: &str,
-    env_var_override_name: &str,
-) -> Result<String> {
-    if let Ok(override_url) = std::env::var(env_var_override_name) {
+async fn maybe_find_peer(role: &ServalRole, override_var: &str) -> Result<String> {
```
shinypb, line 276: Thoughts on initializing the mesh elsewhere and passing a reference to it into this function? (Or does the CLI only ever need to find a peer a single time?)

utils/src/mesh.rs
```diff
@@ -13,6 +14,8 @@ pub trait KaboodleMesh {
 
     /// Create a new entry for a Kaboodle peer network and add ourselves to the mesh.
     async fn start(&mut self) -> Result<(), KaboodleError>;
+    /// Remove this peer from the mesh.
```
shinypb, line 17: Total nit but maybe consider `Remove ourselves from the mesh` so it's more obvious that there is no way to remove forcibly an other peer from the mesh.

cli/src/main.rs
```diff
@@ -320,7 +319,7 @@ async fn main() -> Result<()> {
     SERVAL_NODE_URL.lock().unwrap().replace(baseurl);
 
     match args.cmd {
```
shinypb, line 328: Should we just `.await?` this match statement rather than doing it in all of the arms?

agent/src/main.rs
```diff
@@ -187,7 +187,12 @@ async fn main() -> Result<()> {
         Err(_) => 8181,
     };
 
-    let metadata = PeerMetadata::new(format!("{}:{}", host, port), roles, None);
+    let metadata = PeerMetadata::new(
+        Uuid::new_v4().to_string(),
+        format!("http://{}:{}", host, port),
```
shinypb, line 192: Not a thing for this PR, but it would be ideal to start passing around `SocketAddr`s rather than `host` and `port`, because the way you stringify host is different for IPv4 and IPv6 (which needs `[brackets]` around the IP).

cli/src/main.rs
```diff
@@ -277,27 +285,26 @@ async fn maybe_find_peer(role: &ServalRole, override_var: &str) -> Result<String
...
-        }
+    let candidates = mesh.find_role(role).await;
+    let result = if let Some(target) = candidates.first() {
+        Ok(target.http_address().to_string())
     } else {
         Err(anyhow!("Unable to locate a peer with the {role} role"))
     };
 
-   mesh.stop().await?;
+    mesh.stop().await?;
```
shinypb, line 307: Ok, this finally made it through my skull and brings up a very interesting architectural consideration.

Nothing needs to change right now, but it's worth discussion:

Up until now, I had never contemplated the thought that an instance of the `serval` CLI tool would have to join the mesh to do any work. That totally works, of course, but it seems a bit odd for something so ephemeral to create so much work for the mesh—*everybody* is going to have to update their known peers list, and it will take things a while to settle down—all for a peer who will vanish a few seconds later.

From the perspective of the mesh, a better design would be for the `serval` CLI tool to connect to an instance of `serval-agent` running on localhost. If none is running on localhost, joining the mesh to find a node to talk to would be the fallback. Perhaps we could store all of this in `$TMP` or `/var/run` or whatever?

shinypb (approved):
This looks great. I think it makes sense to move your helpers into the kaboodle repo once we're confident that they're the correct shape (I have no reason to doubt they are, but keeping them here is maximally flexible for the time being).

I don't have any substantial feedback. I have a few friendly nitpicky suggestions and conversation openers below. Shippit!

agent/src/main.rs
```diff
@@ -171,14 +170,34 @@ async fn main() -> Result<()> {
...
 
+    let mesh_port: u16 = match std::env::var("MESH_PORT") {
+        Ok(port_str) => port_str.parse::<u16>().unwrap_or(8181),
+        Err(_) => 8181,
+    };
+
+    let metadata = PeerMetadata::new(
+        Uuid::new_v4().to_string(),
+        format!("http://{}:{}", host, port),
+        roles,
```
shinypb, line 193: It probably doesn’t actually matter, and we certainly don't have to change anything right now, but should we encode the roles in the identity payload as bit masks rather than as strings? The identity payload is included in every single kaboodle packet, so little optimizations add up.

Surely every conceivable role we'd have would fit into a single byte, and since we have versioning (brilliant idea, by the way—I would not have thought of that) we can expand it if we ever need more than 8 roles.

If you agree, I can make a GitHub issue so we don't lose the idea.

cli/src/main.rs
ceejbot: My thinking was that some commands wouldn't need to join the mesh, and joining is expensive in terms of time, and it is one-shot. But if we come across a case where it isn't, then we'd refactor. It's fine either way.

agent/src/api/v1/proxy.rs
ceejbot: It returns an option because the underlying peers() function in kaboodle returns a vec of `(Option<SocketAddr>, Vec<u8>)`!

agent/src/main.rs
ceejbot: :yes:

utils/src/mesh.rs
ceejbot: I renamed this function to `peers_with_role()`, which I think gets across the meaning better.

agent/src/api/v1/proxy.rs
ceejbot: The catastrophic nature of the expect is desirable: this is a crash-and-restart situation. It would be programmer error, most likely.

ceejbot: I will enjoy living to regret this statement.

cli/src/main.rs
ceejbot: No, it would not be functionally equivalent *if* we ever had a non-async command handler. I was preserving the option.

agent/src/main.rs
ceejbot: I did a lot of this work in a commit to this branch today, so I think it's mostly done? I'll review.

agent/src/main.rs
ceejbot: Take a look at how binecode is actually encoding the roles :D It does a very nice job. Well-chosen crate there, colleague!

PR #49, opened 2023-03-24
Enable wasm caching

shinypb:
Per [the conversation in #dev yesterday](https://serval.slack.com/archives/C04BKH1J31S/p1679607830972129), wasmtime's compilation caching is off by default. Enabling it speeds up subsequent runs of the same .wasm file substantially (the ruby example goes from 15.5 seconds to around 0.4 seconds.

I also renamed the parameters to `ServalEngine::execute`, [per this comment](https://github.com/serval/serval-mesh/pull/46#discussion_r1145462026) from yesterday.

### DIFF
--- a/engine/src/lib.rs
+++ b/engine/src/lib.rs
@@ -21,7 +21,7 @@ use wasi_common::{
     pipe::{ReadPipe, WritePipe},
     I32Exit,
 };
-use wasmtime::{Engine, Linker, Module, Store};
+use wasmtime::{Config, Engine, Linker, Module, Store};
 use wasmtime_wasi::{Dir, WasiCtx, WasiCtxBuilder};
 
 pub mod errors;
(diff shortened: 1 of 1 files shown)


engine/src/lib.rs
```diff
@@ -42,7 +42,15 @@ pub struct ServalEngine {
 impl ServalEngine {
     /// Create a new serval engine.
     pub fn new(extensions: HashMap<String, ServalExtension>) -> Result<Self, ServalEngineError> {
-        let engine = Engine::default();
+        let mut config = Config::default();
+        config.cache_config_load_default().map_err(|_| {
```
jkbecker, line 46: Worried about the way wasmtime just intransparently loads some pre-cached binary... Nothing to change here, but I made #51 to follow up on it

jkbecker (approved):
Who doesn't love 50x performance gainz

PR #52, opened 2023-04-04
'just tailscale'

shinypb:
This adds a one-liner `just tailscale` to run serval-agent against your tailnet. It requires `crosswind` to be run separately, but is smart enough to notice when it should suggest that you run crosswind.


ceejbot (approved):
shippit

PR #53, opened 2023-04-06
Mesh shenanigans in the CLI

ceejbot:
Three changes, two notable!

1. Added a new `monitor` command to the cli. This command joins the cli to the mesh as a `client` role, and prints out joins and leaves as it is informed of them.

The CLI and the agent both now respect a new env var: `MESH_INTERFACE`. If you don't provide one, kaboodle defaults to the best it can find. If you do provide a string here, we find an interface with a matching name. This allows me to specify ipv4 and therefore run tests successfully locally.

The CLI now only connects to a mesh to get a base url if it ever needs a base url. It doesn't need a base url for this command, for instance. This change caused a lot of trauma in the cli as things became async that were not previously so. tl;dr the CLI now uses `async_once_cell`.

Generalized some other things as needed. Stole some network interface selection code from kaboodle itself. (We might want to have kaboodle take an interface suggestion via string; worth thinking about anyway.)

2. The CLI now uses the single-peer discovery affordance in the latest Kaboodle. This speeds up all CLI operations _considerably_.

Added new endpoints to the agent to support requesting and receiving lists of peers via the HTTP api. Wrapped kaboodle's new discover-a-peer api in a new `mesh::discover()` function that reuses our env var hinting code. Reworked the CLI's http base url discovery code to use the the new `discover()` function. This code simplified a lot!

`PeerMetadata` and `ServalRole` now both are serializable via Serde.

3. Nuked strum. Strum is nice and it does handy things, but we were only using it for a fairly trivial-to-implement display trait for ServalRole.

### DIFF
--- a/agent/src/api/v1/mesh.rs
+++ b/agent/src/api/v1/mesh.rs
@@ -1,0 +1,30 @@
+use axum::{
+    extract::{Path, State},
+    response::IntoResponse,
+    routing::get,
+    Json,
+};
+use utils::mesh::{KaboodleMesh, PeerMetadata, ServalRole};
+
+use crate::structures::*;
+
+/// Mount all mesh-related introspection endpoints.
+pub fn mount(router: ServalRouter) -> ServalRouter {
+    router
+        .route("/v1/mesh/peers/:role", get(filter_peers)) // TODO
+        .route("/v1/mesh/peers", get(list_peers)) // TODO
+}
+
+/// List all known peers.
+async fn list_peers(_state: State<AppState>) -> Json<Vec<PeerMetadata>> {
+    let mesh = MESH.get().expect("Peer network not initialized!"); // yes, we crash in this case
+    let peers = mesh.peers().await;
+    Json(peers)
+}
+
+/// Filter known peers to only those that advertise the specific role.
+async fn filter_peers(Path(role): Path<ServalRole>, _state: State<AppState>) -> impl IntoResponse {
+    let mesh = MESH.get().expect("Peer network not initialized!"); // yes, we crash in this case
+    let peers = mesh.peers_with_role(&role).await;
+    Json(peers)
+}
--- a/cli/src/main.rs
+++ b/cli/src/main.rs
@@ -8,24 +8,26 @@
 )]
 /// Pounce is a CLI tool that interacts with a running serval agent daemon via
 /// its HTTP API. It discovers running agents via mDNS advertisement.
-use anyhow::{anyhow, Result};
+use anyhow::Result;
 use clap::{Parser, Subcommand};
+use dotenvy::dotenv;
 use humansize::{format_size, BINARY};
 use owo_colors::OwoColorize;
 use prettytable::{row, Table};
-use tokio::time::sleep;
-use utils::mesh::{KaboodleMesh, PeerMetadata, ServalMesh, ServalRole};
 use utils::structs::Manifest;
 use uuid::Uuid;
 
 use std::fs::File;
 use std::io::prelude::*;
 use std::io::BufReader;
-use std::net::SocketAddr;
 use std::path::Path;
 use std::path::PathBuf;
-use std::sync::Mutex;
 use std::time::Duration;
+
+mod mesh;
+mod peers;
+
+use peers::build_url;
 
 #[derive(Parser, Debug)]
 #[clap(name = "pounce 🐈", version)]
@@ -70,21 +72,8 @@ pub enum Command {
     History,
     /// Liveness check: ping at least one node on the mesh.
     Ping,
-}
-
-static SERVAL_NODE_ADDR: Mutex<Option<SocketAddr>> = Mutex::new(None);
-
-/// Convenience function to build urls repeatably.
-fn build_url(path: String, version: Option<&str>) -> String {
-    let baseurl = SERVAL_NODE_ADDR.lock().unwrap();
-    let baseurl = baseurl
-        .as_ref()
-        .expect("build_url called while SERVAL_NODE_URL is None");
-    if let Some(v) = version {
-        format!("http://{baseurl}/v{v}/{path}")
-    } else {
-        format!("http://{baseurl}/{path}")
-    }
+    /// Monitor a mesh: print out new peers and departing peers as we learn about them.
+    Monitor,
 }
 
 async fn upload_manifest(manifest_path: PathBuf) -> Result<()> {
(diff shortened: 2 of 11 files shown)


utils/src/mesh.rs
```diff
@@ -179,3 +193,16 @@ impl KaboodleMesh for ServalMesh {
             .collect()
     }
 }
+
+pub fn mesh_interface_and_port() -> (Option<if_addrs::Interface>, u16) {
+    let mesh_port: u16 = match std::env::var("MESH_PORT") {
```
shinypb, line 215: I just touched this logic earlier this morning and am happy to drop my commit so we don’t conflict, but in case you want it, I thought this was a bit easier to read than our original port parsing code:

````rs
let mesh_port: u16 = std::env::var("MESH_PORT")
    .ok()
    .map(|port_str| port_str.parse().expect("Invalid value given for MESH_PORT"))
    .unwrap_or(8181);
````

File: cli/src/main.rs [hunk 2]
```diff
@@ -111,7 +100,7 @@ async fn upload_manifest(manifest_path: PathBuf) -> Result<()> {
     table.add_row(row!["Wasm task name:", manifest.fq_name()]);
     table.add_row(row!["Version:", manifest.version()]);
 
-    let url = build_url("storage/manifests".to_string(), Some("1"));
+    let url = build_url("storage/manifests".to_string(), Some("1")).await;
```
shinypb, line 103: I was tripping over `build_url` earlier this week; thoughts on a little refactor along these lines? Not asking you to do it necessarily but I just want to run it by you.

```-fn build_url(path: String, version: Option<&str>) -> String {
+fn build_url(path: String) -> String {
+    build_url_with_version(path, Some("1"))
+}
+
+fn build_url_with_version(path: String, version: Option<&str>) -> String {

Also: should we make `path` be a `&str` instead?

File: cli/src/mesh.rs [hunk 3]
```diff
@@ -0,0 +1,49 @@
...
+    let mut mesh = super::peers::create_mesh_peer().await?;
+    let mut discover_rx = mesh
+        .discover_peers()
+        .expect("Unable to get arrivals channel!");
+    let mut depart_rx = mesh
+        .discover_departures()
+        .expect("Unable to get departures channel!");
+
+    loop {
+        if let Some((addr, identity)) = discover_rx.recv().await {
```
shinypb, line 19: Should by a `while let` rather than an `if let`? I think as written, it will only read up to one event from each of the `discover_rx` and `depart_rx` channels and then sleep for 5 seconds, which may be problematic on large meshes.

cli/src/peers.rs
```diff
@@ -0,0 +1,81 @@
...
+        if let Ok(override_addr) = override_url.parse::<SocketAddr>() {
+            return Ok(override_addr);
+        }
+    }
+
+    log::info!("Looking for {role:?} node on the peer network...");
+    let mut mesh = create_mesh_peer().await?;
+
+    // There has to be a better way.
+    sleep(Duration::from_secs(20)).await;
```
shinypb: Do we still need this `sleep`?

cli/src/peers.rs
```diff
@@ -0,0 +1,81 @@
...
+}
+
+async fn maybe_find_peer(role: Option<&ServalRole>, override_var: &str) -> Result<SocketAddr> {
+    if let Ok(override_url) = std::env::var(override_var) {
+        if let Ok(override_addr) = override_url.parse::<SocketAddr>() {
+            return Ok(override_addr);
+        }
+    }
+
+    log::info!("Looking for {role:?} node on the peer network...");
```
shinypb, line 58: Ok, so this is a super interesting function! I have been noodling on this problem myself, and it's how I came up with the "give me a peer with this role" HTTP endpoint.

I was thinking of doing something like this (in terrible pseudocode):

````rs
async fn maybe_find_peer(role: Option<&ServalRole>, override_var: &str) -> Result<KaboodlePeer> {
    let mesh_member = Kaboodle::discover_mesh_member().await
    let Some(required_role) = role else {
        // hey, no required role! that's great, this peer will do
        return mesh_member;
    }

    let http_addr = format!("http://{}", mesh_member.http_address());
    let path = build_url(format!("/mesh/members/{required_role}?count=1"));
    let url = format!("{http_addr}{path}", 

    //  ...and then fetch `url`, parse the response into a KaboodlePeer, return it
}
````

File: cli/src/main.rs (hunk 2 above)
Comment (ceejbot) [reply cli/src/main.rs:103]: Yes to both comments. I'll follow up with this.

File: cli/src/peers.rs (hunk 4 above)
Comment (ceejbot) [reply cli/src/peers.rs]: this is now very very gone. :D

File: cli/src/peers.rs (hunk 5 above)
Comment (ceejbot) [reply cli/src/peers.rs:58]: I like this a bunch as a short-circuit!

File: cli/src/mesh.rs (hunk 3 above)
Comment (ceejbot) [reply cli/src/mesh.rs:19]: We lose departure messages if I do that. Why? I have not thought it through yet. I think there is no reason not to tighten up that loop in the short term, tho.

File: cli/src/peers.rs (hunk 5 above)
Comment (shinypb) [reply cli/src/peers.rs]: Yeah—I think we should never, ever join the mesh unless we're actually there to participate (or be an observer).

File: agent/src/api/v1/mesh.rs [hunk 6]
```diff
@@ -0,0 +1,30 @@
...
+    Json,
+};
+use utils::mesh::{KaboodleMesh, PeerMetadata, ServalRole};
+
+use crate::structures::*;
+
+/// Mount all mesh-related introspection endpoints.
+pub fn mount(router: ServalRouter) -> ServalRouter {
+    router
+        .route("/v1/mesh/peers/:role", get(filter_peers)) // TODO
```
shinypb, line 14: I think these are now `TODONE` 😅

cli/src/mesh.rs
```diff
@@ -43,7 +43,7 @@ pub async fn monitor_mesh() -> anyhow::Result<()> {
         if let Some(addr) = depart_rx.recv().await {
             println!("❌ {} {}", "DEPARTED:".red(), addr);
         }
-        sleep(Duration::from_secs(5)).await;
+        sleep(Duration::from_secs(1)).await;
```
shinypb, line 46: I think we can probably roughly copy this pattern from `KaboodleInner::tick` to avoid the timeout entirely:

````rs
loop {
                tokio::select! {
                    _ = self.broadcast_in_sock.readable() => self.handle_incoming_broadcasts().await,
                    _ = self.sock.readable() => self.handle_incoming_messages().await,
                };
            }
````

PR #55, opened 2023-04-13
Discovery improvements, support for displaying peer latencies

shinypb:
This is a cleaned up version of what I implemented to make the demo video. The three big changes are:

1. pulling in the latest kaboodle, which exposes latency on the `PeerInfo` struct and has faster probe re-broadcasts
2. exposing latency in `serval monitor` — press enter to dump them all out
3. improvements to mesh member discovery: we can now discover mesh members with particular `ServalRole` roles


utils/src/mesh.rs
```diff
@@ -33,7 +37,7 @@ pub trait KaboodlePeer {
 // End of tiny wrapper around Kaboodle.
 
 /// These are the roles we allow peers to advertise on the mesh
-#[derive(Debug, Clone, PartialEq, Eq, Decode, Encode, Deserialize, Serialize)]
+#[derive(Debug, Clone, PartialEq, Eq, Decode, Encode, Hash, Deserialize, Serialize)]
```
shinypb, line 40: All of the derive additions here are so we can stuff these things into a `HashMap`

cli/src/peers.rs
```diff
@@ -11,7 +14,15 @@ static SERVAL_NODE_ADDR: OnceCell<SocketAddr> = async_once_cell::OnceCell::new()
 async fn base_url() -> SocketAddr {
     *SERVAL_NODE_ADDR
         .get_or_init(async {
-            maybe_find_peer("SERVAL_NODE_URL")
+            if let Some(override_addr) = std::env::var("SERVAL_NODE_URL")
```
shinypb, line 17: I moved the env var check up here so I wouldn't have to repeat it in the two very similar `maybe_find_peer` functions

ceejbot: A+; classic refactor

utils/src/structs/api.rs
```diff
@@ -0,0 +1,18 @@
...
+use serde::{Deserialize, Serialize};
+use std::net::SocketAddr;
+
+#[derive(Deserialize, Serialize)]
+pub struct MeshMember {
+    pub http_address: Option<SocketAddr>,
+    pub instance_id: String,
+}
+
+impl From<PeerMetadata> for MeshMember {
```
ceejbot, line 11: Noting out loud: a mesh member is a peer metadata with some metadata removed.

utils/src/mesh.rs
```diff
@@ -176,6 +178,23 @@ impl ServalMesh {
...
+            .filter_map(|(addr, peer_info)| {
+                peer_info.latency.map(|latency| {
+                    (
+                        PeerMetadata::from_identity(addr.ip(), peer_info.identity.to_vec()),
+                        latency,
+                    )
+                })
+            })
+            .collect::<HashMap<_, _>>()
+    }
```
ceejbot, line 196: Check check.

utils/src/mesh.rs
```diff
@@ -78,15 +80,15 @@ struct VersionEnvelope {
     rest: Vec<u8>,
 }
 
-#[derive(Debug, Clone, Deserialize, Serialize)]
+#[derive(Debug, Clone, Eq, PartialEq, Hash, Deserialize, Serialize)]
```
ceejbot, line 83: Since the struct is public, it probably deserves a doc comment on what it is and when to use it.

cli/src/mesh.rs
```diff
@@ -37,7 +42,38 @@ pub async fn monitor_mesh() -> anyhow::Result<()> {
         while let Ok(addr) = depart_rx.try_recv() {
             println!("❌ {} {}", "DEPARTED:".red(), addr);
         }
+        if stdin_rx.try_recv().is_ok() {
```
ceejbot, line 45: we might want a hint to the cli user that pressing enter gets the latencies? I dunno. If you intend on replacing this any time soon leaving it hidden is fine.

ceejbot (approved):
I think we need a little more documentation, but otherwise heck, shippit!

PR #57, opened 2023-04-13
Make jobs runnable even without local storage

ceejbot:
Job runners no longer need local storage. A node with the runner role but without storage now reaches out to a peer with storage to get the executable it's been asked to run.

CAVEATS: There is no attempt made YET to see if a peer node actually has the items we're asking for. So you might try to run a job and hit a peer that doesn't have the thing. We have yet to design how distributing storage will work, so I haven't touched this problem yet.

Here's how this PR achieves this:

The repo has grown a new api-client subcrate. This implements an http client for the serval API. It hides the choice of `reqwest` as an http client-- mostly-- and lets you call async functions to get data.

Re-implemented the cli commands using the api client. While I was there, removed cli support for dead api endpoints and added support for the ones we actually have implemented.

Moved the storage implementation out of utils and into the agent sub-crate, in its own sub-module. The storage implementation is now a trait plus two useful structs that implement the crate. The trait is `RunnerStorage`. This captures what a job runner needs from storage, plus some other not-yet-needed but easy-to-achieve fetches. 

The `cacache`-backed storage struct, `BlobStore`, now implements this trait plus some functions that cacache makes easily available. Notably, getting executable bytes as a *stream* is in the api for this struct and *not* for the trait. I was stuck on wrangling AsyncRead traits, then realized that the job runner doesn't need this, so I punted on it for now.

The second implementor of the RunnerStorage trait is `StorageProxy`. This uses the API client to make requests from a mesh peer that advertises the storage role.

The job runner now uses a lookup function to get its storage implementation. If the node has storage, it snags the `cacache`-backed one it created at startup. If not, it looks up a fresh peer that advertises the role, makes an api client via a StorageProxy, and asks that peer for the data it needs.


agent/src/api/v1/jobs.rs
```diff
@@ -54,15 +54,25 @@ async fn run_job(
...
 
     let Ok(executable) = storage.executable_as_bytes(&name, manifest.version()).await else {
-        return (StatusCode::NOT_FOUND, "no executable found for manifest; key={key}").into_response();
+        return (StatusCode::NOT_FOUND,
+            format!("no executable found for manifest;  name={name}; version={}", manifest.version())).into_response();
     };
 
+    if executable.is_empty() {
+        let warning = format!("Declining to run a job of zero length; name={}; version={}", &name, manifest.version());
+        log::warn!("{warning}");
```
shinypb, line 72: Nit, but `log::warn!` accepts format arguments so you don't actually need to do the `let warning` bit

agent/src/storage/blobs.rs
```diff
@@ -36,51 +44,20 @@ impl BlobStore {
...
+    ) -> Result<ReaderStream<Reader>, ServalError> {
         let key = Manifest::make_executable_key(name, version);
-        let sri = cacache::write(&self.location, key, bytes).await?;
-        Ok(sri.to_string())
+        let fd = cacache::Reader::open(&self.location, key).await?;
+        let stream = ReaderStream::new(fd);
+        Ok(stream)
     }
 
+    // Unused; consider removing
```
shinypb, line 59: I like to put dates in comments like these so the next person to casually read it can have the realization "wow, this has been apparently unused for $long_time" and is more likely to do something about it.

agent/src/storage/blobs.rs
```diff
@@ -130,12 +86,56 @@ impl BlobStore {
...
     }
+}
+
+#[async_trait]
+impl RunnerStorage for &BlobStore {
+    /// Fetch a manifest by its fully-qualified name.
+    async fn manifest(&self, fq_name: &str) -> Result<Manifest, ServalError> {
+        let bytes = cacache::read(&self.location, Manifest::make_manifest_key(fq_name)).await?;
+        if let Ok(data) = String::from_utf8(bytes) {
+            let manifest: Manifest = toml::from_str(&data)?; // TODO CJ this is json right now
```
shinypb, line 97: TODO?

agent/src/storage/blobs.rs
```diff
@@ -130,12 +86,56 @@ impl BlobStore {
...
+        Ok(binary)
+    }
 
-    pub fn manifest_names(&self) -> Result<Vec<String>, ServalError> {
+    /// Retrieve a list of all Wasm manifests stored on this node.
+    async fn manifest_names(&self) -> Result<Vec<String>, ServalError> {
         let result: Vec<String> = cacache::list_sync(&self.location)
             .filter(|xs| xs.is_ok())
             .map(|xs| xs.unwrap().key)
+            .filter(|xs| xs.contains("manifest"))
```
shinypb, line 118: (I am just riffing because coding is fun)

We _could_ smoosh this into a single `.filter_map` rather than `.filter.map.filter`:

````rs
let result: Vec<String> = cacache::list_sync(&self.location)
            .filter_map(|xs| match xs {
                Ok(xs) if xs.key.contains("manifest") => Some(xs.key),
                _ => None,
            })
            .collect();
````

Whether this is a grokkability improvement remains to be seen; I kinda like it more but it's not a slam dunk.

File: justfile [hunk 5]
```diff
@@ -16,7 +16,7 @@ help:
 
 # Run the same checks we run in CI
 @ci: test
-    cargo clippy
+    cargo clippy --workspace
```
shinypb, line 19: Nice

shinypb (approved):
I dig it! Looks like ci is grumpy but I am approving.

agent/src/api/v1/jobs.rs
ceejbot: I re-use the exact same warning text in the next line though :)

agent/src/storage/blobs.rs
ceejbot: I fixed this before the PR; I can remove the TODO!

agent/src/storage/blobs.rs
ceejbot: I kept waiting for clippy to tell me to fix it, but it didn't. I'm meh on the rewrite.

PR #60, opened 2023-04-21
Build releases

jkbecker:
Whipped up a [strongly-inspired-by-tomato](https://github.com/ceejbot/tomato/blob/v0.3.0/.github/workflows/build.yaml) build script. The main difference is using a matrix strategy to reduce the number of steps involved (no need to explicitly write out the build steps for amd64/arm64 separately).

The goal is to have a quick way to install Serval on machines without a Rust toolchain, which is beneficial for agents with low computing power (long build times) and for minimizing dependencies for running Serval in general.

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -1,10 +1,9 @@
 name: Build Serval
 
-on: [push]
-# on:
-#   push:
-#     tags:
-#       - 'v*'
+on:
+  push:
+    tags:
+      - 'v*'
 
 env:
       CARGO_TERM_COLOR: always
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -44,6 +44,9 @@ jobs:
     steps:
       - uses: actions/checkout@v2
 
+      - name: install libssl-dev
+        run: sudo apt-get install -y libssl-dev
+
       - uses: actions-rs/toolchain@v1
         with:
           toolchain: stable
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -1,0 +1,88 @@
+name: Build Serval
+on:
+  push:
+    tags:
+      - 'v*'
+
+env:
+      CARGO_TERM_COLOR: always
+
+jobs:
+  release:
+    name: release
+    runs-on: ubuntu-latest
+    outputs:
+      upload_url: ${{ steps.create_release.outputs.upload_url }}
+    steps:
+      - name: create a github release
+        id: create_release
+        uses: actions/create-release@v1
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        with:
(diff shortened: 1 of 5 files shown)


.github/workflows/build.yaml
```diff
@@ -0,0 +1,154 @@
...
+        arch: [amd64, arm64]
+        include:
+          - arch: amd64
+            target: x86_64-unknown-linux-gnu
+            cross: false
+          - arch: arm64
+            target: aarch64-unknown-linux-gnu
+            cross: true
+      fail-fast: true
+      max-parallel: 1
```
jkbecker, line 45: I noticed that when the matrix runs in parallel, neither of the parallel executions can make use of the cargo cache step. This slows down the overall release creation time but lowers the compute minutes needed to get there.
There may be a way to just have a gatekeeper step for the cache step that lets subsequent ones start up as soon as the cache step of the first variant has completed, but that seems like over-engineering for a first implementation. Unless someone has a very easy 1-liner for this, of course :)

.github/workflows/main.yaml
```diff
@@ -22,8 +22,9 @@ jobs:
       - uses: actions/checkout@v3
       - uses: actions-rs/toolchain@v1
         with:
-          toolchain: stable
+          toolchain: 1.68
```
shinypb, line 25: Can you set up a reminder in slack somewhere so we don’t forget to put this back in a while?

shinypb (approved):
CJ might have feedback but this looks reasonable to me.

Can you write up a quick set of instructions for doing a release and pin it to the #dev channel?

.github/workflows/main.yaml
jkbecker: Yes, will do that. This is really just a bandaid and I kinda hate it

.github/workflows/build.yaml
```diff
@@ -0,0 +1,153 @@
...
+    steps:
+      - uses: actions/checkout@v2
+
+      - name: install libssl-dev
+        run: sudo apt-get install -y libssl-dev
```
ceejbot, line 49: I very much want to reach a point where we don't need this and are using rust's tls implementation throughout. But I get that we need it now.

.github/workflows/build.yaml
```diff
@@ -0,0 +1,153 @@
...
+          key: ${{ runner.os }}-cargo-${{ hashFiles('Cargo.lock') }}
+          restore-keys: |
+            ${{ runner.os }}-cargo-
+
+      - run: rustup target install ${{ matrix.target }}
+        if: ${{ matrix.arch == 'arm'}}
+
+      - run: cargo build --target ${{ matrix.target }} --release
+
+      # TODO: might want to add this to CLI at some point
```
ceejbot, line 132: I'll implement this at some point! It's straightforward :)

ceejbot (approved):
🐑

PR #62, opened 2023-04-25
A storage facade layer + s3 support + CAS endpoint

ceejbot:
This PR significantly reworks the storage layer exposed via the object in the STORAGE once-cell. This is now a single interface, with no exposed traits, that manages all the possible configured storage layers. There are three right now: local storage via cacache, a new AWS S3 bucket, and a proxy that accesses endpoints on mesh peers if the current node has no storage.

The storage functions each implement fallbacks appropriate for each operation. For writes, it will attempt to write to all of its available writable storage. For reads, it prefers local, then tries S3. Users of the storage API (such as the storage endpoints) are not exposed to any of this complexity, as it is hidden away from them.

New crates:
- `aws-sdk-s3` plus `aws-config` for S3 accesses
- `tokio-util` for stream utilities

New env vars respected for configuration:
- `STORAGE_BUCKET`
- `AWS_DEFAULT_REGION`

The AWS configuration will respect all the usual AWS config sources. The simplest way to use this in development is to provide API keys via the two standard env vars.

Extended `ServalError` with more translation types from AWS error types.

The agent http service and the API client library have some new features. Notably, content-addressable storage is now available at `GET /v1/storage/data/*address`. The address must be a valid sub-resource integrity hash, such as
`sha256-Z+FDTdmzupkhgD7AqSKxves583UGzCN/P5xaiu/QL3M=`.

Added data_by_sri() to the storage facade as well as to all the clients. Faked streaming from the proxy client, because I have not yet developed enough streams fu to coerce reqwest's result stream into a more general type.

Cleaned up type signatures for readability. We now have a the type `ServalResult<T>` for returning T plus a `ServalError` from functions.


agent/src/main.rs
```diff
@@ -18,7 +18,7 @@ use axum::extract::DefaultBodyLimit;
 use axum::middleware::{self};
 use axum::routing::get;
 use axum::{Router, Server};
-use dotenvy::dotenv;
+use dotenvy::dotenv_override as dotenv;
```
shinypb, line 21: What does this do?

agent/src/storage/bucket.rs
```diff
@@ -0,0 +1,169 @@
...
+
+#[derive(Debug, Clone)]
+pub struct S3Storage {
+    client: s3::Client,
+    bucket: String,
+}
+
+impl S3Storage {
+    pub fn new(bucket_name: &str, config: aws_config::SdkConfig) -> ServalResult<Self> {
+        let client = s3::Client::new(&config);
```
shinypb, line 16: Asking because I'm still on that learning treadmill: I am surprised to see `config` passed in by reference, rather than handing ownership over. Is that surprising to you, too?

agent/src/storage/bucket.rs
```diff
@@ -0,0 +1,169 @@
...
+        let chunks = object.body.collect().await?;
+        let string = String::from_utf8(chunks.into_bytes().to_vec())?;
+        let manifest: Manifest = toml::from_str(&string)?;
+        Ok(manifest)
+    }
+
+    /// Retrieve a list of all Wasm manifests stored in the bucket.
+    pub async fn manifest_names(&self) -> ServalResult<Vec<String>> {
+        // this is the only complicated one
+        todo!()
```
shinypb, line 68: Noting the todo; but do we really even want/need this function? When would we ever need a complete enumeration? Seems very expensive at scale.

agent/src/storage/mod.rs
```diff
@@ -1,80 +1,348 @@
...
-/// Initialize our local storage.
-pub fn initialize(path: PathBuf) -> Result<(), ServalError> {
-    let store = BlobStore::new(path)?;
+/// Our fully-configured storage object, with all of its details hidden.
+pub static STORAGE: OnceCell<Storage> = OnceCell::new();
+
+/// Initialize our local storage and a proxy option if we have no storage ourselves.
+pub async fn initialize(path: Option<PathBuf>) -> ServalResult<()> {
+    let local = if let Some(blobpath) = path {
+        if let Ok(blob) = BlobStore::new(blobpath) {
```
shinypb, line 34: Noting a bit of discomfort here about tossing away the failure of `BlobStore::new` without doing anything about it. Should we propagate the error up to our caller? Log a warning? I am unsure.

agent/src/storage/mod.rs
```diff
@@ -1,80 +1,348 @@
...
+/// If the operation is a read operation, it tries local storage first then falls back to s3 storage
+/// if that is available. If it's a write operation, it will always try all configured options.
+#[derive(Debug, Clone)]
+pub struct Storage {
+    bucket: Option<S3Storage>,
+    local: Option<BlobStore>,
+}
+
+impl Storage {
+    pub fn new(bucket: Option<S3Storage>, local: Option<BlobStore>) -> Self {
```
shinypb, line 77: I'm sort of viewing Storage as being composed of a hierarchy of storage locations; `local` takes precedence for reads, followed by `bucket`. One can imagine a hypothetical `mesh` option slotting in between them.

agent/src/storage/bucket.rs
ceejbot: tbh the s3 storage should manage creating the config object itself-- it's the only thing that cares.

agent/src/storage/mod.rs
```diff
@@ -1,80 +1,348 @@
...
+            return proxy.store_manifest(manifest).await;
+        }
+
+        let local_result = if let Some(local) = &self.local {
+            Some(local.store_manifest(manifest).await)
+        } else {
+            None
+        };
+
+        let bucket_result = if let Some(bucket) = &self.bucket {
```
shinypb, line 178: I may be trying to be too fancy, but can this be done as a `map` to avoid the if/else? As in: I am still learning Rust, literally can we even do this? If not, why not? "It's not idiomatic Rust" is a very useful response.

(Computers are fun! This is intended as a friendly conversation, not a nitpick.)

````rs
let bucket_result = &self.bucket.map(|bucket| bucket.store_manifest(manifest).await)```

File: agent/src/storage/mod.rs [hunk 7]
```diff
@@ -1,80 +1,348 @@
...
+            Some(bucket.store_manifest(manifest).await)
+        } else {
+            None
+        };
+
+        // Consider comparing integrity hashes.
+
+        if let Some(result) = local_result {
```
shinypb, line 186: Similar code golfing thought here; `.or` might let us avoid this if/else block:

```let result = local_result.or(bucket_result).unwrap_or_else(|| Err(...))```

shinypb (approved):
LGTM! Kicking the tires on my machine now, but merge away

agent/src/storage/mod.rs
ceejbot: The async-ness of the call makes this a little exciting.

PR #63, opened 2023-04-27
S3 buckets now support fetching content by address

ceejbot:
Added a layer of indirection to S3 storage as a cheap index. Instead of writing data to the user-provided key, we write an object containing the integrity hash for the data. We then create an object that contains the data using the integrity hash as a key. Re-implemented all data stores and fetches in the bucket via this indirection. This is of course invisible to clients of the storage layer.

Implemented `HEAD /v1/storage/data/*address` via `has_content_address()`, which is now possible that the bucket storage supports it.

The term sri is now gone from all API surfaces. Instead we use "integrity" as the word for this concept.

Improved some error reporting, and fixed a bug with how the cli reported hashes. (It was reporting hashes of the hashes instead of printing the hashes, because I held the API wrong.)

### DIFF
--- a/agent/src/storage/bucket.rs
+++ b/agent/src/storage/bucket.rs
@@ -2,6 +2,7 @@ use aws_sdk_s3 as s3;
 use s3::error::ProvideErrorMetadata;
 use s3::primitives::ByteStream;
 use ssri::Integrity;
+use urlencoding::encode;
 use utils::errors::{ServalError, ServalResult};
 
 #[derive(Debug, Clone)]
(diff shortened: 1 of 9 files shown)


agent/src/storage/bucket.rs
```diff
@@ -20,39 +21,69 @@ impl S3Storage {
...
+        &self,
+        integrity: &Integrity,
+        bytes: &[u8],
+    ) -> ServalResult<Integrity> {
+        let body = ByteStream::from(bytes.to_vec());
+        let result = self
+            .client
+            .put_object()
+            .bucket(&self.bucket)
+            .key(encode(&integrity.to_string())) // integrity string is not url-safe!
```
shinypb, line 47: It's a real bummer that the AWS client doesn't just take care of this for us automatically.

agent/src/storage/mod.rs
```diff
@@ -90,6 +90,42 @@ impl Storage {
     // This implementation is just a bunch of painful by-hand delegation logic.
     // I'd like to golf it down.
 
+    /// Store a blob of data in the content-addressable store, responding with the
+    /// integrity hash of the data.
+    pub async fn store_by_integrity(&self, bytes: &[u8]) -> ServalResult<Integrity> {
+        if !self.has_storage() {
+            // let proxy = make_proxy_client().await?;
+            // return proxy.store_executable(name, version, bytes.to_vec()).await;
+            todo!();
```
shinypb, line 99: Helpfully calling out this `todo!`

shinypb:
Looks like there's a formatting lint that's failing, but this looks great to me. I am excited. :+1: Ship when ready

Issue #61, opened 2023-04-21
Regression with clap and Rust 1.69


jkbecker:
Since the release of Rust 1.69, our clippy CI check throws this error and fails:
```
error: this looks like you are trying to swap `__clap_subcommand` and `Subcommand`
  --> cli/src/main.rs:44:24
   |
44 | #[derive(Clone, Debug, Subcommand)]
   |                        ^^^^^^^^^^
   |
   = note: or maybe you should use `std::mem::replace`?
   = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#almost_swapped
note: the lint level is defined here
  --> cli/src/main.rs:44:24
   |
44 | #[derive(Clone, Debug, Subcommand)]
   |                        ^^^^^^^^^^
   = note: `#[deny(clippy::almost_swapped)]` implied by `#[deny(clippy::correctness)]`
   = note: this error originates in the derive macro `Subcommand` (in Nightly builds, run with -Z macro-backtrace for more info)

error: could not compile `serval` due to previous error
warning: build failed, waiting for other jobs to finish...
error: could not compile `serval` due to previous error
```

As a quick-fix bandaid, #60 pinned the Rust toolchain to 1.68 in the following places:
- a newly introduced [rust-toolchain.toml](https://github.com/serval/serval-mesh/pull/60/files#diff-2b1bde2cf3a858b7bf7424cb8bcbf01f35b94dc80b925d9432cbab3319ca9b4e)
- the [main.yaml](https://github.com/serval/serval-mesh/pull/60/files#diff-71cabc4177e41ea8f15a89eb65398fa8187739f68a2762706d84885cd8b2ddc3) general CI checks
- the [build.yaml](https://github.com/serval/serval-mesh/pull/60/files#diff-d0777657fa3fd81d23aaf7273e58aee453b04e67882517900c56daeef9b3e4c1) release builder

This helps make CI not fail in the short run, but is obviously not a good solution.

Creating this issue and as per [this remark](https://github.com/serval/serval-mesh/pull/60#discussion_r1174023550)

jkbecker:
Fixed in #64

PR #64, opened 2023-04-28
chore: update clap to 4

ceejbot:
We no longer have colorized help. At some point we'll follow up with whatever clap settles on for styling help output.

### DIFF
--- a/cli/Cargo.toml
+++ b/cli/Cargo.toml
@@ -8,7 +8,7 @@ license = "BSD-2-Clause-Patent"
 anyhow = { workspace = true }
 async-once-cell = "0.4.4"
 atty = { workspace = true  }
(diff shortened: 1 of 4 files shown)


cli/Cargo.toml
```diff
@@ -8,7 +8,7 @@ license = "BSD-2-Clause-Patent"
 anyhow = { workspace = true }
 async-once-cell = "0.4.4"
 atty = { workspace = true  }
-clap = { version = "3.2.23", features = ["derive", "wrap_help"] } # older version on purpose
+clap = { version = "4.2.4", features = ["derive", "wrap_help"] } # older version on purpose
```
jkbecker, line 11: the "older version..." comment can go

jkbecker (changes requested):
if you un-pin the toolchain as outlined in #61 we're good to go :)

PR #65, opened 2023-04-28
The start of an implementation of a job scheduler

shinypb:
This is incomplete in a few obvious ways; in particular, priority levels other than `Normal` and distribution models / kinds other than `OneOff` are not supported. I am sure I will also find plenty of other missing things when I start writing the API endpoints that will invoke all of this stuff. However! It seemed good to get the basic implementation in front of another human being's brain before I go on for too much longer.

Critiques on style, implementation, and overall direction are welcome. The matchmaking logic here is currently extremely basic, assigning jobs to runners on a first-come, first-serve basis. I suspect we'll need something more clever in the not-so-distant future, but let's see how far we can get with the stupidest possible implementation.

### DIFF
--- a/agent/src/scheduler/mod.rs
+++ b/agent/src/scheduler/mod.rs
@@ -1,0 +1,514 @@
+#![allow(dead_code)] // temporary, during initial development
+
+use std::collections::HashMap;
+use std::time::{Duration, SystemTime};
+
+/// https://www.notion.so/srvl/Job-Scheduler-spec-6f8860f3e6874341aba0b286373d5f67?pvs=4
+///
+/// The scheduler, formerly known as the job queue, is in charge of creating scheduled jobs and
+/// shepherding them through their lifecycle. At a high level, a scheduled job consists of:
+/// - a pointer to a Manifest in the storage system (via an ssri::Integrity hash)
+/// - a pointer to an input payload in the storage system
+/// - various configuration values indicating how and where the job should be executed
+///
+/// The scheduler is intended to be intelligent about where it runs jobs, sending them to the most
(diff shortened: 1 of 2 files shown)


agent/src/scheduler/mod.rs
```diff
@@ -0,0 +1,508 @@
...
+enum ScheduledJobRequirement {
+    /// Requires that the runner has the given extension available.
+    Extension(String),
+    /// Requires that the runner has `/proc` (e.g. is Linux)
+    Proc,
+}
+
+/// Represents the kind of invocation this job would like to receive.
+#[derive(Debug, PartialEq)]
+enum ScheduledJobKind {
```
shinypb, line 91: These are referred to as "Distribution Models" in the spec, and as "Kind" here. I don't like either of these terms. @ceejbot I am begging you to help me come up with a good name for this concept.

agent/src/scheduler/mod.rs
```diff
@@ -0,0 +1,508 @@
...
+    /// This job should be scheduled with a normal amount of priority.
+    #[default]
+    Normal = 2,
+    /// This job should only be scheduled if there is nothing more important to do.
+    LowPriority = 3,
+}
+
+/// Represents a constraint on the runner that can run a job.
+#[derive(Debug, PartialEq)]
+enum ScheduledJobRequirement {
```
shinypb, line 82: Name feedback welcome here as well; this enum is used in two places:

- inside of the `ScheduledJob` struct to indicate which capabilities the job requires
- inside of the `available_runners` HashMap to indicate which capabilities a given runner can offer

PR #67, opened 2023-04-28
A PATCH endpoint for the content-addressable store

ceejbot:
`utils::diffs::make_patch()` and `utils::diffs::make_patch()` now exist to make compressed binary patches. The intended use case is to send executable deltas around instead of full payloads. These use the new dependency [qbsdiff](https://lib.rs/crates/qbsdiff) to generate and apply compressed binary patches.

Implemented `PATCH /v1/storage/data/*address` using the `apply_patch()` function. It's up to you
to generate the patch that gets applied somehow. (qbsdiff has command-line tools for this.)

Renamed some storage functions to make it clear that they return streams, to distinguish them from ones that return vecs of u8.


utils/src/diffs.rs
```diff
@@ -0,0 +1,53 @@
...
+
+pub fn make_patch(source: &[u8], target: &[u8]) -> ServalResult<Vec<u8>> {
+    let mut patch = Vec::new();
+    Bsdiff::new(source, target).compare(Cursor::new(&mut patch))?;
+    Ok(patch)
+}
+
+pub fn apply_patch(source: &[u8], patch: &[u8]) -> ServalResult<Vec<u8>> {
+    let patcher = Bspatch::new(patch)?;
+    let mut target = Vec::with_capacity(patcher.hint_target_size() as usize);
```
shinypb, line 15: First time seeing a hint actually get used. Fun! Rust feels like Real Software Engineering™ sometimes. 😅

agent/src/api/v1/storage.rs
```diff
@@ -136,6 +138,44 @@ async fn has_content_address(Path(address): Path<String>) -> impl IntoResponse {
...
+    let Some(storage) = STORAGE.get() else {
+        return (StatusCode::SERVICE_UNAVAILABLE, "storage uninitialized; programmer error".to_string()).into_response();
+    };
+
+    let Ok(integrity) = address.parse::<Integrity>() else {
+        let e = ServalError::BlobAddressInvalid(format!("{} is not a valid sub-resource integrity string", address));
+        return e.into_response()
+    };
+
+    match storage.data_by_integrity(integrity).await {
```
shinypb, line 152: Just noting for my own edification: I think the implication of this is that we can currently only patch things that will fit in memory, but that seems like a problem that won't bite us any time soon.

shinypb:
You have no idea how excited I am about this.

agent/src/api/v1/storage.rs
ceejbot: This limitation is absolutely real for us in several ways right now.

PR #68, opened 2023-05-01
A meta endpoint serving hostname (for now) + more

jkbecker:
A `/meta` endpoint serving (for now) the agent's hostname.
This is the most basic implementation; other metadata could/should probably be contained in here as well.

### DIFF
--- a/agent/src/api/mod.rs
+++ b/agent/src/api/mod.rs
@@ -1,9 +1,11 @@
 use anyhow::Result;
 use axum::extract::State;
+use gethostname::gethostname;
 use axum::http::{Request, StatusCode};
 use axum::middleware::Next;
 use axum::response::{IntoResponse, Response};
 use http::header::HeaderValue;
+use serde::Serialize;
 
 use crate::structures::AppState;
 
@@ -51,4 +53,18 @@ pub async fn ping() -> String {
 pub async fn monitor_status(_state: State<AppState>) -> impl IntoResponse {
     metrics::increment_counter!("monitor:status");
     StatusCode::NOT_IMPLEMENTED
+}
+
+#[derive(Serialize)]
+pub struct AgentMetadata {
+    hostname: String,
+}
+
+/// Provide agent metadata.
+pub async fn meta() -> String {
+    metrics::increment_counter!("meta");
+    let meta = AgentMetadata {
+        hostname: gethostname().into_string().expect("Failed to get hostname"),
+    };
+    serde_json::to_string(&meta).expect("Failed to serialize agent metadata")
 }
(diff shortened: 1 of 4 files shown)


agent/src/main.rs
```diff
@@ -236,7 +236,8 @@ fn init_router(state: &Arc<RunnerState>) -> Router {
 
     let mut router: Router<Arc<RunnerState>, Body> = Router::new()
         .route("/monitor/ping", get(ping))
-        .route("/monitor/status", get(monitor_status));
+        .route("/monitor/status", get(monitor_status))
+        .route("/meta", get(meta));
```
shinypb, line 240: Can we stick this under something like `/meta/hostinfo`? I want the /meta/ namespace to be open to future expansion.

ceejbot: `/status` is the name I chose because it should include a lot more than host info. Here's an example response blob that contains some of the information I consider essential for lightweight runtime inspection of a process. 

```json
{
   "git": "b2277c7",
    "hostname": "unobtanium.local",
    "node": "name-of-agent",
    "stats": {
        "requestCount": 0,
        "statuses": {}
    },
    "uptime": 22.108189125
}```

Plus whatever else is cheap to extract from a running process: peer counts? etc. Uptime is particularly useful to me. PID if we have it. (these two together can signal crash loops.)

Comment (ceejbot) [reply agent/src/main.rs:240]: (Getting the git hash involves a little build-time Rust fun that will show you how all that works.)

File: agent/src/api/mod.rs [hunk 2]
```diff
@@ -52,3 +54,17 @@ pub async fn monitor_status(_state: State<AppState>) -> impl IntoResponse {
...
+    hostname: String,
+}
+
+/// Provide agent metadata.
+pub async fn meta() -> String {
+    metrics::increment_counter!("meta");
+    let meta = AgentMetadata {
+        hostname: gethostname().into_string().expect("Failed to get hostname"),
+    };
+    serde_json::to_string(&meta).expect("Failed to serialize agent metadata")
```
ceejbot, line 69: Building a response this way works around how axum expects to serve json-- you'll lose the content-type header. Follow the example in the axum docs for [how to serve json](https://docs.rs/axum/latest/axum/struct.Json.html#response-example).

agent/src/main.rs
jkbecker: Thx, great input! Will add the very easy stats details to my PR and create separate issues for those that are more complex.

jkbecker:
what about a route of `/agent/info`  for the functionality above?
that would allow for clean extension into basically self-documenting endpoints like
```
/agent/neighbors
/agent/load
/agent/extensions
```
…?

PR #70, opened 2023-05-25
Update wasmtime from v5.0.0 to v9.0.1

ceejbot:
This required removing the experimental wasi http crates. Git remembers all if we should need to re-do this work.

### DIFF
--- a/engine/src/lib.rs
+++ b/engine/src/lib.rs
@@ -23,7 +23,6 @@ use wasmtime_wasi::{Dir, WasiCtx, WasiCtxBuilder};
 pub mod errors;
 pub mod extensions;
 mod runtime;
-use wasi_experimental_http_wasmtime::{HttpCtx, HttpState};
 
 use crate::errors::ServalEngineError;
 use crate::runtime::register_exports;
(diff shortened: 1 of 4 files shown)


engine/src/lib.rs
```diff
@@ -78,32 +77,6 @@ impl ServalEngine {
         let stdout = WritePipe::new_in_memory();
         let stderr = WritePipe::new_in_memory();
 
-        // Link the experimental HTTP support
```
shinypb, line 81: Consider removing the [HTTP-related permissions](https://github.com/serval/serval-mesh/blob/c76a430924c0008a9f30c3ea1ed9de06ff97bd6f/utils/src/structs/mod.rs#L238-L239) and the [http-test sample project](https://github.com/serval/wasm-samples/tree/main/http-test).

shinypb (approved):
🫡
🌊🌊🌊🌊🌊🌊
🌊⛴️🌊🌊🌊🌊
🌊🌊🌊🐟🌊🌊
🪸🪸🪸🪸🪸🪸

